PR #66, opened 2026-04-13
fix: quarantine mechs with unreachable IPFS tools manifests

LOCKhart07:
## Summary

- A single mech with a broken metadata CID (e.g. `0x33ca1e117c4254b2ee8cd7ef1621739431a37396` → CID `f01701220…` returning `protobuf: (PBNode) invalid wireType` from the IPFS gateway) previously caused `MechInformationRound` to return `None` on every attempt, cycling the FSM back to `MechVersionDetectionRound` indefinitely.
- `populate_tools` now quarantines individual mechs once the tools-manifest retries are exhausted, skips them on subsequent iterations, and lets the round complete with the remaining mechs.
- `get_mechs_info` only returns `None` when **every** mech has failed (so the FSM still retries cleanly rather than persisting an all-empty batch). A per-round `_failed_mechs` set is cleared in `clean_up`, so a republished CID is retried next time the round runs.
- Operators retain the existing `ignored_mechs` skill param as the permanent escape hatch for persistently broken mechs; the log now calls out quarantined addresses explicitly to make that signal easy to spot.

## Test plan

- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py -rfE` — 13 passed (5 new, 8 existing)
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests -rfE` — 331 passed
- [x] `uv run tomte check-code` — black, isort, flake8, mypy, pylint, darglint all OK
- [x] `make generators` — skill hash updated in `packages.json` / `skill.yaml`
- [ ] Smoke test on mainnet: confirm logs show quarantine of `0x33ca1e11…` and the FSM advances past `MechInformationRound` instead of cycling

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
@@ -98,8 +98,13 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
 
             if len(res) == 0:
                 self.context.logger.warning(
-                    f"The {mech.address} mech agent's tools are empty!"
+                    f"Quarantining mech {mech.address}: tools manifest at "
+                    f"{self.mech_tools_api.url} is empty. Empty lists are "
+                    f"deterministic per-CID; will retry on next round entry."
                 )
+                self._failed_mechs.add(mech.address)
+                self.mech_tools_api.reset_retries()
+                continue
 
             # store only the relevant mech tools
             relevant_tools = set(res) - self.params.irrelevant_tools
--- a/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
@@ -21,7 +21,7 @@
 """This module contains the behaviour responsible for gathering information about the mech marketplace."""
 
 import json
-from typing import Any, Generator, Optional
+from typing import Any, Generator, Optional, Set
 
 from packages.valory.skills.mech_interact_abci.behaviours.base import (
     MechInteractBaseBehaviour,
@@ -54,6 +54,7 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
         """Initialize Behaviour."""
         super().__init__(**kwargs)
         self._fetch_status: FetchStatus = FetchStatus.NONE
+        self._failed_mechs: Set[str] = set()
 
     @property
     def mech_tools_api(self) -> MechToolsSpecs:  # pragma: no cover
@@ -73,7 +74,7 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
     ) -> WaitableConditionType:
         """Populate the tools of the mech info, using the metadata."""
         for mech in mech_info:
-            if mech.relevant_tools:
+            if mech.relevant_tools or mech.address in self._failed_mechs:
                 continue
 
             self.set_mech_agent_specs(mech.service.metadata_str)
@@ -85,6 +86,14 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
                 msg = f"Could not get the {mech.address} mech agent's tools from {self.mech_tools_api.url}."
                 self.context.logger.warning(msg)
                 self.mech_tools_api.increment_retries()
+                if self.mech_tools_api.is_retries_exceeded():
+                    self.context.logger.error(
+                        f"Quarantining mech {mech.address}: could not fetch "
+                        f"tools manifest from {self.mech_tools_api.url} "
+                        f"after retries exhausted."
+                    )
+                    self._failed_mechs.add(mech.address)
+                    self.mech_tools_api.reset_retries()
                 return False
 
             if len(res) == 0:
@@ -113,17 +122,20 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
             return None
 
         while True:
-            # fails even if a single mech's set of tools cannot be fetched
             tools_populated = yield from self.populate_tools(mech_info)
             if tools_populated:
                 break
 
-            if not self.mech_tools_api.is_retries_exceeded():
-                continue
+        if self._failed_mechs:
+            self.context.logger.warning(
+                f"Skipped {len(self._failed_mechs)} mech(s) with unreachable "
+                f"tools manifests: {sorted(self._failed_mechs)}"
+            )
 
-            msg = "Retries were exceeded while trying to get the mech tools."
-            self.context.logger.warning(msg)
-            self.mech_tools_api.reset_retries()
+        if not any(mech.relevant_tools for mech in mech_info):
+            self.context.logger.warning(
+                "No mechs have usable tools after fetch; emitting NONE to retry."
+            )
             return None
 
         # truncate the information, otherwise logs get too big
@@ -151,3 +163,4 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
     def clean_up(self) -> None:  # pragma: no cover
         """Clean up the behaviour."""
         self.mech_tools_api.reset_retries()
+        self._failed_mechs.clear()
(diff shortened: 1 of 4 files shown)


packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -85,6 +86,14 @@ def populate_tools(
...
+                    self.context.logger.error(
+                        f"Quarantining mech {mech.address}: could not fetch "
+                        f"tools manifest from {self.mech_tools_api.url} "
+                        f"after retries exhausted."
+                    )
+                    self._failed_mechs.add(mech.address)
+                    self.mech_tools_api.reset_retries()
                 return False
 
             if len(res) == 0:
```
bennyjo, line 99: The quarantine path only triggers when `process_response` returns `None`. If a mech's manifest resolves HTTP-wise but yields `[]` consistently, `mech.relevant_tools` stays empty, the mech is re-fetched on every `populate_tools` pass, and `get_mechs_info` returns `None` via the `not any(mech.relevant_tools ...)` check — same cycling symptom as the reported bug, different trigger.

Is the `0x33ca1e11…` CID guaranteed to always hit the `res is None` branch? If empty-list is possible, this path should also feed the quarantine counter, otherwise the fix only closes the observed trigger.

bennyjo (commented):
Scope question on the empty-tools path.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
```diff
@@ -281,3 +283,173 @@ def mock_fetch_mechs_info():
...
+
+
+class TestQuarantine:
+    """Tests for per-mech quarantine on IPFS fetch failure."""
+
+    def test_populate_tools_quarantines_mech_after_retries_exhausted(self) -> None:
+        """A mech whose fetch fails with retries exhausted is added to _failed_mechs."""
+        behaviour = _make_mech_info_behaviour()
+        behaviour._context.params = MagicMock()
+        behaviour._context.params.ipfs_address = "https://ipfs.io/"
```
bennyjo, line 297: Two coverage gaps on an urgent fix:

1. All new tests set `is_retries_exceeded.return_value = True`, so quarantine fires on the first failure. The production flow — `populate_tools` returns `False` several times while retries build, then quarantines on the Nth call — isn't exercised. A test flipping `is_retries_exceeded` from `False` to `True` after N calls would cover the `while True` multi-iteration path.
2. No negative test asserting that a single failure with `is_retries_exceeded=False` does **not** add the mech to `_failed_mechs`. Without it, `test_populate_tools_quarantines_mech_after_retries_exhausted` would still pass if the guard were `if True:`.

bennyjo (commented):
Test coverage gaps worth closing before merge.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
LOCKhart07: Good catch — you're right that empty-list was a distinct cycling vector. Verified the reported CID (`0x33ca1e11…`) returns HTTP 500 + plain-text error, so it does hit the `res is None` path and the original quarantine fires. But a mech publishing `{"tools": []}` would indeed keep every mech in `relevant_tools = set()` and trip the `not any(...)` guard → `None` → cycle.

Fixed in 637aa4f: `len(res) == 0` now also adds the mech to `_failed_mechs` (immediate, no retries — empty is deterministic per-CID). `clean_up` clears the set, so a republished manifest gets retried on the next round entry. Unit test at `test_quarantines_on_empty_tools` covers the path.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
LOCKhart07: Both gaps addressed in 2f27b18:

1. **Multi-iteration build-up** → `test_quarantine_only_fires_on_final_retry`: `is_retries_exceeded` flips from `False` to `True` after N `increment_retries` calls. `populate_tools` is invoked in a loop mirroring `get_mechs_info`'s outer `while True`. Asserts `_failed_mechs` stays empty through iterations 1..N-1 and only fills on the final pass, and that `increment_retries` was called exactly N times.

2. **Negative test** → `test_populate_tools_does_not_quarantine_before_retries_exceeded`: single failure with `is_retries_exceeded.return_value = False`. Asserts `_failed_mechs == set()` and `reset_retries` was *not* called. Falsifiable: if the `is_retries_exceeded` guard were removed, this test fails immediately.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -85,12 +86,25 @@ def populate_tools(
...
                 self.mech_tools_api.increment_retries()
+                if self.mech_tools_api.is_retries_exceeded():
+                    self.context.logger.error(
+                        f"Quarantining mech {mech.address}: could not fetch "
+                        f"tools manifest from {self.mech_tools_api.url} "
+                        f"after retries exhausted."
+                    )
+                    self._failed_mechs.add(mech.address)
+                    self.mech_tools_api.reset_retries()
                 return False
```
DIvyaNautiyal07, line 97: **Non-blocking (future improvement):** No backoff between retries — when `populate_tools` returns `False`, the `while True` loop in `get_mechs_info` immediately retries with no delay. For permanent content errors (like the protobuf wireType failure), all 6 attempts are wasted in rapid succession.

`ApiSpecs` already exposes a `suggested_sleep_time` property (exponential backoff via `backoff_factor`). Adding a sleep here would reduce IPFS request hammering and leave more of the 30s round timeout budget for healthy mechs:

```python
yield from self.sleep(self.mech_tools_api.suggested_sleep_time)
return False
```

Also worth considering: reducing `retries: 5` to `2` or `3` in skill.yaml — for permanent content errors, fewer wasted attempts means faster quarantine.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -85,12 +86,25 @@ def populate_tools(
                 msg = f"Could not get the {mech.address} mech agent's tools from {self.mech_tools_api.url}."
                 self.context.logger.warning(msg)
                 self.mech_tools_api.increment_retries()
+                if self.mech_tools_api.is_retries_exceeded():
```
DIvyaNautiyal07, line 89: **Non-blocking (future improvement):** `process_response` returns `None` for both transient errors (IPFS gateway unreachable/timeout) and permanent errors (malformed protobuf content that will never parse). The quarantine mechanism treats them identically — retry N times, then give up.

A future improvement could override `MechToolsSpecs.process_response` to return a richer signal (e.g. distinguish JSON parse failure from HTTP error), enabling immediate quarantine on permanent content errors with 0 wasted retries.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -151,3 +168,4 @@ def async_act(self) -> Generator:  # pragma: no cover
     def clean_up(self) -> None:  # pragma: no cover
         """Clean up the behaviour."""
         self.mech_tools_api.reset_retries()
+        self._failed_mechs.clear()
```
DIvyaNautiyal07, line 171: **Non-blocking (future consideration):** `_failed_mechs` is instance state cleared on every round transition — including `ROUND_TIMEOUT`. For the reported single-mech incident this is fine (~6s of retries, well within the 30s timeout). But if multiple mechs have broken CIDs and their combined retry time exceeds the 30s round timeout, the round times out, quarantine state is lost, and the next run starts fresh — same loop.

If this becomes an issue, persisting quarantined addresses in `SynchronizedData` (with a TTL or generation counter) would survive timeout resets.

DIvyaNautiyal07 (commented):
Reviewed the fix against the reported incident (single mech with malformed IPFS metadata cycling the FSM indefinitely). The quarantine logic is correct and sufficient for the reported scenario. Three non-blocking suggestions for future hardening are left as inline comments.

DIvyaNautiyal07 (approved):

bennyjo (approved):
All blocking review comments addressed: empty-list quarantine vector closed in 637aa4f, test coverage gaps closed in 2f27b18 with falsifiable multi-iteration and negative tests. Non-blocking suggestions are appropriate follow-up work.

PR #68, opened 2026-04-13
feat: classify permanent mech-tools errors for immediate quarantine

LOCKhart07:
## Summary

Follow-up to PR #66 addressing Divya's [non-blocking review comment](https://github.com/valory-xyz/mech-interact/pull/66#discussion_r3074198125) on transient-vs-permanent error conflation.

Before this change, `populate_tools` treated every failed IPFS fetch as potentially transient — increment the retry counter, quarantine only after 6 attempts. For the reported `0x33ca1e117c4254b2ee8cd7ef1621739431a37396` incident (HTTP 500 + `protobuf: (PBNode) invalid wireType`) this wasted ~1.8s of the round's time budget retrying bytes that would never change.

`MechToolsSpecs.is_permanent_error(response)` now classifies:
- **4xx** → permanent (gateway rejects the request — 404 CID not found, 400 bad request, etc.)
- **2xx with failed `process_response`** → permanent (server returned stable malformed content; retries won't change it)
- **5xx with body markers** → permanent if body contains any of `protobuf:`, `invalid wiretype`, `malformed`, `failed to resolve`, `cid not found` (case-insensitive, substring match)
- **Everything else** → transient (default skews safe; preserves the existing retry path for genuine gateway flakes like 502/503/504)

`populate_tools` branches on the classifier: permanent → straight into `_failed_mechs` with no `increment_retries`; transient → existing retry path unchanged.

## Impact

| Scenario | Before | After |
|---|---|---|
| 1 broken mech (reported incident) | 2.1s | ~0.35s |
| 5 broken mechs | ~10.5s | ~1.75s |
| Round-timeout ceiling (permanent errors, 30s budget) | ~14 mechs | ~85 mechs |

Fix 4 (parallelization, PR #69) takes this to effectively unbounded.

## Test plan

- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/test_models.py::TestIsPermanentError -rfE` — 16 classifier unit tests (protobuf body, gateway timeout, 404, 502, 200+bad-JSON, 200+missing-key, case-insensitivity, non-UTF8 bytes, all markers)
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py -rfE` — 18 tests (3 new for permanent branch, 15 existing adapted to pin classifier mock)
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests -rfE` — 352 passed total
- [x] `uv run tomte check-code` — black, isort, flake8, mypy, pylint, darglint all OK
- [x] `make generators` — skill hash updated
- [x] **Real trader2 smoke test against `0x33ca1e11…`** — log at `2026-04-14 00:58:42` shows single `Could not get` warning → immediate `Quarantining mech 0x33ca1e11…: permanent content error at … (status=500); retries skipped.` → `Event.DONE`. No retries. Confirmed working end-to-end.

## Files changed

- `packages/valory/skills/mech_interact_abci/models.py` — `MechToolsSpecs.is_permanent_error` + marker constants
- `packages/valory/skills/mech_interact_abci/behaviours/mech_info.py` — classifier branch in `populate_tools`
- `packages/valory/skills/mech_interact_abci/tests/test_models.py` — 16 classifier tests
- `packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py` — 3 new branch tests + 5 adapted existing tests
- `packages/packages.json`, `skill.yaml` — hash update

## Complements

- PR #66 (per-mech quarantine) — merged; this builds on it.
- PR #69 (parallelize fetches) — separate; takes the ceiling to unbounded. Stacked on this branch.

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
@@ -85,6 +85,17 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
             if res is None:
                 msg = f"Could not get the {mech.address} mech agent's tools from {self.mech_tools_api.url}."
                 self.context.logger.warning(msg)
+
+                if self.mech_tools_api.is_permanent_error(res_raw):
+                    self.context.logger.error(
+                        f"Quarantining mech {mech.address}: permanent content "
+                        f"error at {self.mech_tools_api.url} "
+                        f"(status={res_raw.status_code}); retries skipped."
+                    )
+                    self._failed_mechs.add(mech.address)
+                    self.mech_tools_api.reset_retries()
+                    return False
+
                 self.mech_tools_api.increment_retries()
                 if self.mech_tools_api.is_retries_exceeded():
                     self.context.logger.error(
--- a/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
@@ -69,6 +69,12 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
         self.mech_tools_api.url = ipfs_link
         self.mech_tools_api.__dict__["_frozen"] = True
 
+    def _quarantine_mech(self, mech_address: str, reason: str) -> None:
+        """Log the quarantine, mark the mech failed, and reset the retry counter."""
+        self.context.logger.error(f"Quarantining mech {mech_address}: {reason}")
+        self._failed_mechs.add(mech_address)
+        self.mech_tools_api.reset_retries()
+
     def populate_tools(
         self, mech_info: MechsSubgraphResponseType
     ) -> WaitableConditionType:
@@ -87,24 +93,20 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
                 self.context.logger.warning(msg)
 
                 if self.mech_tools_api.is_permanent_error(res_raw):
-                    self.context.logger.error(
-                        f"Quarantining mech {mech.address}: permanent content "
-                        f"error at {self.mech_tools_api.url} "
-                        f"(status={res_raw.status_code}); retries skipped."
+                    self._quarantine_mech(
+                        mech.address,
+                        f"permanent content error at {self.mech_tools_api.url} "
+                        f"(status={res_raw.status_code}); retries skipped.",
                     )
-                    self._failed_mechs.add(mech.address)
-                    self.mech_tools_api.reset_retries()
                     return False
 
                 self.mech_tools_api.increment_retries()
                 if self.mech_tools_api.is_retries_exceeded():
-                    self.context.logger.error(
-                        f"Quarantining mech {mech.address}: could not fetch "
-                        f"tools manifest from {self.mech_tools_api.url} "
-                        f"after retries exhausted."
+                    self._quarantine_mech(
+                        mech.address,
+                        f"could not fetch tools manifest from "
+                        f"{self.mech_tools_api.url} after retries exhausted.",
                     )
-                    self._failed_mechs.add(mech.address)
-                    self.mech_tools_api.reset_retries()
                 return False
 
             if len(res) == 0:
--- a/packages/valory/skills/mech_interact_abci/models.py
+++ b/packages/valory/skills/mech_interact_abci/models.py
@@ -57,34 +57,54 @@ class MechToolsSpecs(ApiSpecs):
     # Substrings (case-insensitive) in a 5xx response body that indicate the
     # CID is structurally invalid and will never resolve. 5xx without any
     # marker is treated as transient (gateway flake).
-    PERMANENT_ERROR_BODY_MARKERS: ClassVar[Tuple[str, ...]] = (
+    _PERMANENT_5XX_BODY_MARKERS: ClassVar[Tuple[str, ...]] = (
         "invalid wiretype",
-        "protobuf:",
         "malformed",
         "failed to resolve",
         "cid not found",
     )
+
+    # Only the first slice of a 5xx body is scanned for permanent markers:
+    # gateway error lines always appear at the top, and unbounded decode of
+    # hostile/misconfigured multi-MB bodies is wasteful.
+    _PERMANENT_5XX_BODY_SCAN_BYTES: ClassVar[int] = 4096
 
     def is_permanent_error(self, response: HttpMessage) -> bool:
         """Classify a failed response as permanent (never retry) or transient.
 
         Called by populate_tools only when process_response returned None.
         Permanent means quarantine immediately; transient means defer to the
-        retry counter.
+        retry counter. Caller guarantees `response` is non-None and
+        `response.status_code` is a populated int: the framework contract for
+        `get_http_response` returns a fully-formed `HttpMessage` or raises
+        (transport failures raise `TimeoutException`, they do not synthesize a
+        partial message).
+
+        2xx is treated as permanent because `process_response` currently only
+        returns None on a 2xx when the body is malformed content (not a flake).
+        If `process_response` is ever widened to signal other failure modes on
+        2xx, this branch must be revisited.
 
         :param response: the HTTP response whose body/status is classified.
         :return: True if the error is permanent and retries should be skipped.
         """
         status = response.status_code
         if 200 <= status < 300:
-            # 2xx + process_response failed => malformed content, not a flake.
             return True
         if 400 <= status < 500:
             # Gateway rejected the request (404 CID not found, etc.).
             return True
         if 500 <= status < 600:
-            body = response.body.decode(errors="ignore").lower()
-            return any(marker in body for marker in self.PERMANENT_ERROR_BODY_MARKERS)
+            body = (
+                response.body[: self._PERMANENT_5XX_BODY_SCAN_BYTES]
+                .decode(errors="ignore")
+                .lower()
+            )
+            return any(marker in body for marker in self._PERMANENT_5XX_BODY_MARKERS)
+        self.context.logger.warning(
+            f"Unclassified failure status {status} for {self.url}; "
+            f"treating as transient. Investigate if this recurs."
+        )
         return False
 
 
--- a/packages/valory/skills/mech_interact_abci/models.py
+++ b/packages/valory/skills/mech_interact_abci/models.py
@@ -59,8 +59,6 @@ class MechToolsSpecs(ApiSpecs):
     # marker is treated as transient (gateway flake).
     _PERMANENT_5XX_BODY_MARKERS: ClassVar[Tuple[str, ...]] = (
         "invalid wiretype",
-        "malformed",
-        "failed to resolve",
         "cid not found",
     )
 
--- a/packages/valory/skills/mech_interact_abci/models.py
+++ b/packages/valory/skills/mech_interact_abci/models.py
@@ -20,7 +20,7 @@
 """This module contains the models for the abci skill of MechInteractAbciApp."""
 
 from dataclasses import dataclass
-from typing import Any, Dict, FrozenSet, List, Optional, cast
+from typing import Any, ClassVar, Dict, FrozenSet, List, Optional, Tuple, cast
 
 from aea.exceptions import enforce
 from aea.skills.base import SkillContext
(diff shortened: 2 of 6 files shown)


packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -85,6 +85,17 @@ def populate_tools(
             if res is None:
                 msg = f"Could not get the {mech.address} mech agent's tools from {self.mech_tools_api.url}."
                 self.context.logger.warning(msg)
+
+                if self.mech_tools_api.is_permanent_error(res_raw):
```
bennyjo, line 95: **Open for discussion — potentially blocking.**

`is_permanent_error(res_raw)` dereferences `res_raw.status_code` and `res_raw.body` directly. Pre-PR, `res_raw` was only passed to `process_response`, which presumably tolerated `None` on transport failures (connection refused, timeout). If `get_http_response` can yield `None` here, this raises `AttributeError` in `async_act` — turning a previously-transient failure into a round-level crash on exactly the failure mode this PR targets.

Is `res_raw` framework-guaranteed non-None? If yes, worth a docstring line on `is_permanent_error` pinning the invariant + a test. If no, a `if res_raw is None: return False` guard at the top of the classifier.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -85,6 +85,17 @@ def populate_tools(
...
+                    self.context.logger.error(
+                        f"Quarantining mech {mech.address}: permanent content "
+                        f"error at {self.mech_tools_api.url} "
+                        f"(status={res_raw.status_code}); retries skipped."
```
bennyjo: **Open for discussion — related to the comment on line 89.**

`(status={res_raw.status_code})` assumes `status_code` is a populated int. If the framework ever yields a response object with `status_code = None` (e.g. connection reset mid-stream), every such failure classifies as transient via the 500-range check — silently reverting to the old retry-wastage behaviour for exactly the failure mode this PR is meant to fix.

Worth an explicit branch: *no response or no status → transient; otherwise classify*.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -54,6 +54,39 @@
...
+        """
+        status = response.status_code
+        if 200 <= status < 300:
+            # 2xx + process_response failed => malformed content, not a flake.
+            return True
+        if 400 <= status < 500:
+            # Gateway rejected the request (404 CID not found, etc.).
+            return True
+        if 500 <= status < 600:
+            body = response.body.decode(errors="ignore").lower()
```
bennyjo: **Optional polish — feel free to defer.**

A few things worth considering on the classifier body, none blocking:

- **Body size**: `response.body.decode(errors="ignore").lower()` materialises + lowercases the full body. A misconfigured or hostile gateway returning a multi-MB 5xx causes an O(N) allocation per failing fetch. Slicing to the first 4KB before `.lower()` is enough — permanent markers are always in the first error line. A compiled `re.IGNORECASE` regex over the markers also removes the `.lower()` and the five `in` scans in one go.
- **Marker breadth**: `"protobuf:"` as a substring matches inside unrelated diagnostic text (e.g. `"using protobuf: v3 — upstream timeout"`) and lets a hostile gateway inject markers to force quarantine. Tighter anchoring (leading word-boundary or exact phrase with adjacent punctuation) mitigates both.
- **2xx coupling**: 2xx + failed `process_response` → permanent is correct today but couples this classifier to `process_response`'s failure modes. A docstring line pinning that contract would prevent a future widening of `process_response` from silently causing premature quarantine.
- **3xx fallthrough**: 3xx with `process_response == None` almost certainly means a misconfigured redirect, not a flake. A `logger.warning` on unclassified status would catch future gateway drift instead of burning retries silently.
- **Naming**: `PERMANENT_ERROR_BODY_MARKERS` is only consulted in the 5xx branch — consider `_PERMANENT_5XX_BODY_MARKERS` to match usage.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -85,6 +85,17 @@ def populate_tools(
...
+                    )
+                    self._failed_mechs.add(mech.address)
+                    self.mech_tools_api.reset_retries()
+                    return False
```
bennyjo, line 101: **Optional polish.**

The "res is None" block now has two quarantine paths + one retry path, each with their own logger + `_failed_mechs.add` + `reset_retries`. A small `_quarantine(mech, reason)` helper would remove ~10 lines of duplication and stop the two branches drifting on bookkeeping.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
```diff
@@ -546,3 +552,148 @@ def test_clean_up_clears_failed_mechs(self) -> None:
 
         assert behaviour._failed_mechs == set()
         mock_api.reset_retries.assert_called_once()
+
+
```
bennyjo, line 556: **Optional polish — test coverage gaps in `TestPermanentErrorClassification`:**

- No test for `res_raw is None` → classifier path (tied to the open discussion on `mech_info.py:89`).
- No test for transient-then-permanent on the same mech (retry counter incremented once, then gateway returns a permanent body — `reset_retries` is called in the permanent branch, but nothing pins that).
- `test_get_mechs_info_permanent_error_needs_only_one_http_call` encodes traversal order via `process_response.side_effect = [["tool_good"], None]`. If `fetch_mechs_info` ever reorders, the test silently tests the wrong thing — mapping by URL would be more robust.

bennyjo (commented):
Scope matches the reported incident. Two open questions on network-error handling (comments on mech_info.py:89 and :93); the rest are optional polish.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -54,6 +54,39 @@
...
     """A model that wraps ApiSpecs for the Mech agent's tools specifications."""
 
+    # Substrings (case-insensitive) in a 5xx response body that indicate the
+    # CID is structurally invalid and will never resolve. 5xx without any
+    # marker is treated as transient (gateway flake).
+    PERMANENT_ERROR_BODY_MARKERS: ClassVar[Tuple[str, ...]] = (
+        "invalid wiretype",
+        "protobuf:",
+        "malformed",
+        "failed to resolve",
```
DIvyaNautiyal07: **P2: `"failed to resolve"` is ambiguous.** IPFS gateways use this phrasing both for structurally invalid CIDs (permanent) and for CIDs where no connected peer currently serves the content (transient — may recover). The other markers (`protobuf:`, `invalid wiretype`) are unambiguous; this one isn't. A valid-but-unserved CID would be quarantined for the round when a retry might have recovered.

Consider narrowing to `"failed to resolve path"` or requiring co-occurrence with a structural marker.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -54,6 +54,39 @@
 class MechToolsSpecs(ApiSpecs):
     """A model that wraps ApiSpecs for the Mech agent's tools specifications."""
 
+    # Substrings (case-insensitive) in a 5xx response body that indicate the
+    # CID is structurally invalid and will never resolve. 5xx without any
+    # marker is treated as transient (gateway flake).
+    PERMANENT_ERROR_BODY_MARKERS: ClassVar[Tuple[str, ...]] = (
+        "invalid wiretype",
+        "protobuf:",
+        "malformed",
```
DIvyaNautiyal07: **P2: `"malformed"` is a broad substring match.** Any 5xx body containing "malformed" anywhere classifies as permanent (e.g. a gateway intermediary returning "malformed X-Forwarded-For header"). Low likelihood for IPFS GETs in practice, but less specific than the sibling markers.

Consider `"malformed object"` / `"malformed CID"` if you can pin down the actual gateway string.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
```diff
@@ -546,3 +552,148 @@ def test_clean_up_clears_failed_mechs(self) -> None:
...
+        # Classifier was invoked on the received response.
+        mock_api.is_permanent_error.assert_called_once_with(http_message)
+        # Error log mentions permanent content error for ops telemetry.
+        error_calls = behaviour.context.logger.error.call_args_list
+        assert any("permanent content error" in str(call) for call in error_calls)
+
+    def test_transient_error_still_increments_retries_and_does_not_quarantine(
+        self,
+    ) -> None:
+        """Transient path is byte-identical to pre-Fix-2 behaviour."""
```
DIvyaNautiyal07: **Nit:** "Fix 2" is an internal revision label that won't mean anything to a reader post-merge. Same at lines 642-643 and 697. Consider `"pre-classifier behaviour"` / `"Before this change"`.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
LOCKhart07: Verified the framework contract: `get_http_response` is typed `Generator[None, None, HttpMessage]` (`abstract_round_abci/behaviour_utils.py:1291`) — it returns an `HttpMessage` or raises `TimeoutException`. Transport failures raise, they don't synthesize a partial message. Also worth noting: `process_response` itself already dereferences `response.body.decode()` with no None guard (`abstract_round_abci/models.py:719`), so if `res_raw` could be None the pre-PR code on line 83 would already crash.

Rather than planting unreachable defensive code (which contradicts the "trust framework guarantees" rule), I've pinned the invariant in the `is_permanent_error` docstring so future readers don't re-discover this. Happy to revisit if you'd prefer the guard.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
LOCKhart07: Same framework-contract analysis as the line-89 reply: `get_http_response` returns a fully-formed `HttpMessage` (real int `status_code`) or raises `TimeoutException`; there's no code path that yields `status_code=None`. A mid-stream reset raises, it doesn't return a half-built message.

I've extended the `is_permanent_error` docstring to also pin the `status_code: int` invariant. One related point you raised does apply though — 1xx/3xx fallthrough — and I've addressed that separately: unclassified statuses now emit a `logger.warning` before returning transient, so future gateway drift surfaces in ops logs.

packages/valory/skills/mech_interact_abci/models.py
LOCKhart07: Applied the full set:

- **Body size:** now slices to `_PERMANENT_5XX_BODY_SCAN_BYTES = 4096` before `.decode().lower()`. Gateway error lines always appear in the first line, so 4KB is generous, and hostile multi-MB bodies are bounded. Falling through to transient (when a marker somehow lives past 4KB) is the safe direction — the retry counter still quarantines eventually.
- **Marker breadth:** dropped `"protobuf:"` entirely. `"invalid wiretype"` is the actual protobuf-decode signature we care about, so no recall loss, and the injection surface is gone. (Stayed with substring matching rather than a compiled regex — the scan is now bounded to 4KB so the perf case for regex disappears.)
- **2xx coupling:** added a docstring paragraph pinning the `process_response`-failure contract so a future widening would flag this branch.
- **3xx fallthrough:** unclassified status now emits a `logger.warning` before returning transient.
- **Naming:** renamed to `_PERMANENT_5XX_BODY_MARKERS` (plus `_PERMANENT_5XX_BODY_SCAN_BYTES`).

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
LOCKhart07: Extracted `_quarantine_mech(mech_address, reason)` on `MechInformationBehaviour`. Both quarantine sites now route through it, so `logger.error` + `_failed_mechs.add` + `reset_retries` have a single source of truth and can't drift.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
LOCKhart07: - **`res_raw is None` test:** skipped. Per the framework-contract analysis in the line-89 thread, there's no reachable code path that produces `None` here; a test would either construct an impossible scenario or test Python's attribute model rather than our code. The invariant is now pinned in the `is_permanent_error` docstring.
- **Transient-then-permanent:** added `test_transient_then_permanent_resets_retries_on_quarantine`. It runs `populate_tools` twice on the same mech — call 1 transient (pins `increment_retries` called, `reset_retries` not), call 2 permanent (pins `reset_retries` is called in the permanent branch even after a prior transient bumped the counter).
- **Traversal-order side_effect:** `process_response.side_effect` is now a callable keyed on `mock_api.url` (set dynamically by `set_mech_agent_specs` from each mech's `metadata_str`). The good mech always populates and the broken mech always classifies on its first attempt regardless of iteration order.

DIvyaNautiyal07 (approved):

packages/valory/skills/mech_interact_abci/models.py
LOCKhart07: Good catch — agreed this is ambiguous. Dropped `"failed to resolve"` entirely in b47ce04 rather than narrowing. Rationale: the structural-CID case is already caught by `"invalid wiretype"` and `"cid not found"` on 5xx, plus the unconditional 4xx branch (the autonolas gateway's real CID-invalid body is `"invalid path "...": invalid cid: trailing bytes..."` which arrives as 400). No recall loss; removes the ambiguity; errs toward retry-then-quarantine on the no-peers-serving case.

packages/valory/skills/mech_interact_abci/models.py
LOCKhart07: Dropped `"malformed"` entirely in b47ce04. On checking: we have no real-world incident body that contains it — the only 5xx we've actually observed is `"failed to resolve /ipfs/...: protobuf: (PBNode) invalid wireType..."` (caught by `"invalid wiretype"`), and the autonolas gateway's CID-structural errors come through as 400 with `"invalid cid: trailing bytes..."` (already unconditionally permanent). `"malformed"` was speculative, so narrowing it would just be speculation with extra steps; removing it is the cleaner call.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
LOCKhart07: Fixed in b47ce04 — all three references now use descriptive wording ("Transient path still runs through the retry counter unchanged" / "Without the classifier this would be 7 ..." etc.).

bennyjo (approved):
Ready to merge. Classifier is conservatively biased toward transient, markers are tightened to the two evidence-backed signatures, 4KB body scan bounds hostile decode, and tests pin the cross-branch reset_retries bookkeeping. The 2xx-is-permanent branch is load-bearing on the documented process_response invariant — worth revisiting if that surface ever widens, but non-blocking today.

Issue #67, opened 2026-04-13
Parallelize populate_tools IPFS fetches for restart resilience


LOCKhart07:
## Context

Even with [PR #66](https://github.com/valory-xyz/mech-interact/pull/66) (per-mech quarantine) and a follow-up to classify permanent vs transient errors landed, a fresh agent restart facing many malformed mech manifests can still time out `MechInformationRound`. `_failed_mechs` is instance state and is empty on every restart. Each broken mech costs ≥1 HTTP round-trip sequentially, so N broken mechs cost N × RTT. At a 30s round timeout and ~400ms typical RTT, the ceiling is ~75 broken mechs — plausible in the wild if a popular mech operator publishes a bad CID.

Observed in a trader2 run against PR #66: `0x33ca1e117c4254b2ee8cd7ef1621739431a37396` took 2.1s of sequential retries to quarantine; round finished at ~15s. Linear scaling in the number of broken mechs.

SynchronizedData persistence of `_failed_mechs` (tracked separately) only helps once a round successfully commits a payload — if every fresh restart times out before `Event.DONE`, nothing persists. Parallelization is the primary lever for restart resilience.

**Goal:** fire all outstanding IPFS manifest fetches concurrently and await them together. Wall time becomes `max(per-mech RTT)` instead of `sum(per-mech RTT)`.

### Why this is harder than it looks

`BaseBehaviour.get_http_response` is a blocking generator (`abstract_round_abci/behaviour_utils.py:1284-1320`) built on `_do_request` → `wait_for_message`. The `wait_for_message` docstring explicitly says _\"does not handle concurrent requests\"_ (`behaviour_utils.py:276`). There is no `asyncio.gather`-style primitive in the Valory skill stack.

However, the underlying infrastructure does support multiple in-flight requests:

- `Requests` model (`abstract_round_abci/models.py:547-556`) is `request_id_to_callback: Dict[str, Callable]` — one entry per in-flight request, keyed by `request_nonce`.
- `AbstractResponseHandler.handle` (`handlers.py:354-429`) dispatches incoming messages by nonce and invokes the registered callback — multiple coexist fine.
- `self.context.outbox.put_message` is non-blocking.

The sugar layer is what serialises. Bypass it: send N messages, register a callback per nonce that stashes the response in a dict, poll until all responded or a budget expires.

### Proposed primitive

Add a helper to `MechInformationBehaviour` (not the base class — keep blast radius small):

\`\`\`python
[...]

PR #69, opened 2026-04-13
feat: parallelize populate_tools IPFS fetches

LOCKhart07:
## Summary

Fix 4 from the follow-up plan on PR #66. Parallelizes the IPFS manifest fetches in `MechInformationBehaviour.populate_tools` so broken mechs don't serialise a fresh agent restart into a round timeout.

**Based on PR #68** (permanent-error classifier) — this PR targets that branch. After #68 merges, I'll rebase onto main.

Problem: PR #66 gave us per-mech quarantine; PR #68 collapsed permanent-error quarantine to 1 attempt. Both still run sequentially — N broken mechs take N × RTT wall-clock. Observed 2.1s for the reported `0x33ca1e11…` case; ceiling at 30s round timeout is ~85 permanently-broken mechs with PR #68 alone. After this PR: effectively unbounded (wall time = `max(per-mech RTT)`).

## Approach

`BaseBehaviour.get_http_response` is sequential by design — `wait_for_message` explicitly forbids concurrent requests. But the underlying infrastructure is concurrency-friendly: `Requests.request_id_to_callback` is nonce-keyed (`secrets.token_hex(32)` — never collides), `AbstractResponseHandler` dispatches by nonce, `put_message` is non-blocking.

Added a local helper `_fetch_http_parallel` on `MechInformationBehaviour` that:

1. Calls `_build_http_request_message` N times (unique nonce per dialogue).
2. Registers **custom** nonce-scoped callbacks (not `get_callback_request` — that routes through `try_send` which is serial and state-dependent).
3. Puts all N messages to outbox up front.
4. Polls via `self.sleep` until every callback has fired or a timeout budget expires.
5. Cleans up unresolved nonces from the registry in `finally` so late responses don't keep closures alive across rounds.

The custom callback is dead-simple and behaviour-state-agnostic:

```python
def _make_parallel_fetch_callback(results, nonce):
    def _callback(message, current_behaviour):
        results[nonce] = message
    return _callback
```

## Semantics changes worth highlighting

- **Shared retry counter advances once per batch pass**, not per-mech. On exhaustion, all still-unpopulated mechs are quarantined together.
- **Timed-out fetches** (None response) are treated as transient; increment the shared counter, quarantine only on retry exhaustion.
- **Specs are snapshotted per mech** before the shared URL is mutated for the next mech.

Full design/rationale: `/home/lockhart/.claude/plans/mech-info-parallel-fetch.md` (local). Issue #67 has a thinner earlier draft (now obsolete — see this PR).

## New skill param

`mech_tools_parallel_timeout: 20.0` (seconds) — leaves 10s of the 30s round budget for subgraph fetch + consensus payload. **Optional with a 20.0s default** (`kwargs.get`, not `_ensure`), so downstream skills inheriting `MechParams` (trader_abci, market_manager_abci, etc.) don't need to touch their `skill.yaml` to pick up this fix.

## Commits

- `e194766` — `_fetch_http_parallel` primitive + 6 isolation tests.
- `60c51ae` — `populate_tools` rewrite + 6 parallel-semantics tests + adapted existing tests.
- `060fc1b` — fix: make `mech_tools_parallel_timeout` optional so downstream skills load without config changes.

## Test plan

- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py::TestFetchHttpParallel -rfE` — 6 tests: fan-out, input-order return, timeout-returns-None, registry cleanup on timeout, registry cleanup on exception, custom callback isolation.
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py::TestPopulateToolsParallelSemantics -rfE` — 6 tests: mixed-outcome batch, single-counter-advance per pass, batch-quarantine on exhaustion, timeout-as-transient, zero-pending early return, per-mech spec snapshot.
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests -rfE` — 364 passed.
- [x] `uv run tomte check-code` — all 6 linters green.
- [x] `make generators` — skill hash updated.
- [x] **Real trader2 smoke test against `0x33ca1e11…`** (log `2026-04-14 00:58:38→00:58:42`):
  - `mech_information_round` → `Event.DONE` in ~4s (dominated by subgraph fetch upstream).
  - Single `Could not get` warning, single `Quarantining mech 0x33ca1e11…: permanent content error … retries skipped.` No retries.
  - FSM advances normally past `mech_information_round` → `fetch_markets_router_round` → `update_bets_round` → etc.
- [x] **Downstream skill load** — trader_abci loads without needing `mech_tools_parallel_timeout` in its `skill.yaml` (verified after `060fc1b`).
- [ ] Synthetic N-broken-mech run — not executed (the `0x33ca1e11…` case is the only permanently-broken mech observed; we'd need to stand up bogus CIDs to stress it).

## References

- Infrastructure trace: `abstract_round_abci/behaviour_utils.py:1259-1398`, `handlers.py:354-489`, `dialogues.py:99-125`, `models.py:547-556`.
- PR #66 (per-mech quarantine) — merged.
- PR #68 (permanent-error classifier) — this PR's base.
- Issue #67 — tracking issue; spec now obsolete (replaced by the design above and the local plan file).

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
@@ -21,8 +21,12 @@
 """This module contains the behaviour responsible for gathering information about the mech marketplace."""
 
 import json
-from typing import Any, Generator, Optional, Set
+import time
+from typing import Any, Callable, Dict, Generator, List, Optional, Set
 
+from aea.protocols.base import Message
+
+from packages.valory.protocols.http import HttpMessage
 from packages.valory.skills.mech_interact_abci.behaviours.base import (
     MechInteractBaseBehaviour,
     WaitableConditionType,
@@ -43,6 +47,25 @@ from packages.valory.skills.mech_interact_abci.states.mech_info import (
 )
 
 CID_PREFIX = "f01701220"
+PARALLEL_FETCH_POLL_INTERVAL = 0.1
+
+
+def _make_parallel_fetch_callback(
+    results: Dict[str, Optional[HttpMessage]], nonce: str
+) -> Callable[[Message, Any], None]:
+    """Build a nonce-scoped callback that stores the response in results.
+
+    Stays behaviour-state-agnostic so it dispatches regardless of whether the
+    owning behaviour is in WAITING_MESSAGE state (we aren't — we poll via
+    self.sleep while the fetch is in flight).
+    """
+
+    def _callback(message: Message, current_behaviour: Any) -> None:
+        # The mech_interact parallel fetcher only ever receives HttpMessage
+        # responses for these nonces; the cast is safe by construction.
+        results[nonce] = message  # type: ignore[assignment]
+
+    return _callback
 
 
 class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
@@ -55,6 +78,8 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
         super().__init__(**kwargs)
         self._fetch_status: FetchStatus = FetchStatus.NONE
         self._failed_mechs: Set[str] = set()
+        # Injectable clock for deterministic parallel-fetch timeout tests.
+        self._clock: Callable[[], float] = time.monotonic
 
     @property
     def mech_tools_api(self) -> MechToolsSpecs:  # pragma: no cover
@@ -68,6 +93,45 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
         self.mech_tools_api.__dict__["_frozen"] = False
         self.mech_tools_api.url = ipfs_link
         self.mech_tools_api.__dict__["_frozen"] = True
+
+    def _fetch_http_parallel(
+        self,
+        specs: List[Dict[str, Any]],
+        timeout: float,
+        poll_interval: float = PARALLEL_FETCH_POLL_INTERVAL,
+    ) -> Generator[None, None, List[Optional[HttpMessage]]]:
+        """Fire N HTTP requests concurrently; return responses in input order.
+
+        :param specs: list of kwargs dicts for _build_http_request_message.
+        :param timeout: wall-clock budget (via self._clock) for the whole batch.
+        :param poll_interval: sleep between readiness checks.
+        :return: list with one entry per input spec; None for timed-out
+            requests. Unresolved callbacks are popped from the request
+            registry on exit.
+        :yield: None while awaiting responses.
+        """
+        results: Dict[str, Optional[HttpMessage]] = {}
+        nonces: List[str] = []
+        registry = self.context.requests.request_id_to_callback
+
+        for spec in specs:
+            message, dialogue = self._build_http_request_message(**spec)
+            nonce = self._get_request_nonce_from_dialogue(dialogue)
+            nonces.append(nonce)
+            registry[nonce] = _make_parallel_fetch_callback(results, nonce)
+            self.context.outbox.put_message(message=message)
+
+        deadline = self._clock() + timeout
+        try:
+            while len(results) < len(nonces):
+                if self._clock() >= deadline:
+                    break
+                yield from self.sleep(poll_interval)
+        finally:
+            for nonce in nonces:
+                registry.pop(nonce, None)
+
+        return [results.get(n) for n in nonces]
 
     def populate_tools(
         self, mech_info: MechsSubgraphResponseType
--- a/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
@@ -136,20 +136,44 @@ class MechInformationBehaviour(QueryingBehaviour, MechInteractBaseBehaviour):
     def populate_tools(
         self, mech_info: MechsSubgraphResponseType
     ) -> WaitableConditionType:
-        """Populate the tools of the mech info, using the metadata."""
-        for mech in mech_info:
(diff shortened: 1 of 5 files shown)


packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -69,43 +94,96 @@ def set_mech_agent_specs(self, metadata: str) -> None:
...
+            # get_spec() BEFORE advancing to the next mech so each sent
+            # request carries its own mech's CID.
             self.set_mech_agent_specs(mech.service.metadata_str)
-            specs = self.mech_tools_api.get_spec()
-            res_raw = yield from self.get_http_response(**specs)
+            specs_per_mech.append(dict(self.mech_tools_api.get_spec()))
+
+        responses = yield from self._fetch_http_parallel(
+            specs_per_mech,
+            timeout=self.params.mech_tools_parallel_timeout,
```
bennyjo, line 190: `self.mech_tools_api.url` has been mutated to the last mech's URL by the time this loop runs — the spec snapshot loop above leaves it pointing at `pending[-1]`. Every warning here will report the wrong URL for all but the last mech in the batch. Same concern for any other log in this response loop that reads `self.mech_tools_api.url`.

Suggest logging the snapshotted URL instead:

```suggestion
            if res_raw is None:
                self.context.logger.warning(
                    f"Timed out fetching {mech.address} mech agent's tools "
                    f"from {specs_per_mech[pending.index(mech)]['url']}."
                )
```

(or zip `specs_per_mech` alongside `pending`/`responses` to avoid the `.index()` lookup.)

bennyjo (commented):
One finding — stale URL in warning log.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -69,43 +94,96 @@ def set_mech_agent_specs(self, metadata: str) -> None:
...
+    def _fetch_http_parallel(
+        self,
+        specs: List[Dict[str, Any]],
+        timeout: float,
+        poll_interval: float = PARALLEL_FETCH_POLL_INTERVAL,
+    ) -> Generator[None, None, List[Optional[HttpMessage]]]:
+        """Fire N HTTP requests concurrently; return responses in input order.
+
+        :param specs: list of kwargs dicts for _build_http_request_message.
+        :param timeout: wall-clock budget (via self._clock) for the whole batch.
```
DIvyaNautiyal07, line 117: **P2 — Partial-registration leak on fan-out failure.** The fan-out loop registers callbacks in `registry` before the `try/finally` begins. If `_build_http_request_message`, `_get_request_nonce_from_dialogue`, or `put_message` raises on iteration N, the first N-1 nonces stay in `request_id_to_callback` with closures over `results`, surviving across rounds (the `finally` at line ~105 never runs because the try hasn't been entered).

`test_cleanup_runs_on_exception` only covers `self.sleep` raising, not a fan-out exception. Framework methods don't raise in normal operation, so this is defensive-only, but the fix is cheap: move the `for spec in specs:` loop inside the `try:` block so the existing `finally` cleanup covers partial registration too.

DIvyaNautiyal07 (approved):

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
LOCKhart07: Good catch — fixed in 3aa674e (merge commit dc6a7c1). The response loop now zips in `specs_per_mech` and logs `mech_url = spec.get("url", ...)` so each warning reports the correct per-mech URL instead of the last-mutated `self.mech_tools_api.url`. Applied to both the timeout-warning and process-failure-warning paths, and propagated into the `_quarantine_mech` reason strings as well.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
LOCKhart07: Fair point — fixed in 3aa674e (merge commit dc6a7c1). Moved the `for spec in specs:` fan-out loop inside the `try:` block in `_fetch_http_parallel`, so the existing `finally` now also cleans up any nonces registered before a mid-loop raise. Added a comment in the code explaining why registration lives inside the try.

You're right that `test_cleanup_runs_on_exception` only covered `self.sleep` raising — the structural fix is so cheap it didn't seem worth a dedicated mid-fan-out raise test, but happy to add one if you'd prefer.

bennyjo (approved):
Solid change. Parallel primitive is cleanly isolated and the wall-time-vs-sum test is genuinely falsifiable. Two small follow-ups worth considering (non-blocking): (1) reset the shared retry counter on the zero-pending early return in populate_tools so every exit path is symmetric; (2) add a short comment above _fetch_http_parallel's try: naming the BaseBehaviour invariants the parallel path depends on (nonce-keyed registry, non-blocking put_message, nonce-based handler dispatch) as a tripwire for future refactors.

LOCKhart07:
Thanks for the review @bennyjo! Both follow-ups applied in 539c811:

1. **Symmetric retry-counter reset**: added `self.mech_tools_api.reset_retries()` on the zero-pending early return in `populate_tools`. In practice retries is always 0 at that point (end-of-pass either resets or exhausts-then-resets), but the explicit call is a tripwire — future refactors that change end-of-pass bookkeeping won't silently leave retries mid-count on this path.

2. **Invariants comment on `_fetch_http_parallel`**: added a block above the `try:` naming the three framework invariants the parallel path depends on — nonce-keyed `request_id_to_callback` dict, non-blocking `outbox.put_message` (multiplexer schedules asynchronously), and nonce-based handler dispatch (independent of behaviour state). Verified each against the open-autonomy/open-aea source before writing.

PR #74, opened 2026-04-29
revert(mech_info): drop parallel fetch, coalesce by CID instead

LOCKhart07:
## Summary

Reverts the parallel `populate_tools` path introduced in #69 and replaces the lost throughput recovery with **per-CID coalescing**.

A trader running mech-interact v0.26.0 inside Pearl wedged silently inside `mech_information_round` on Omen (4 mechs sharing one metadata CID — `0x157d3b10…`). Reverting the trader to v0.25.0 (pre-#69) restored normal operation. v0.26.0's only silent code path is #69's `_fetch_http_parallel` custom-callback dispatch, so the regression localizes there.

## Why parallelize was speculative

#67's motivation was a hypothetical N-broken-mechs ceiling. The actually-observed worst case in production was a single broken mech (~2.1s sequentially). #68's permanent-error classifier already collapses broken-mech cost to one RTT, so the realistic ceiling sits around ~75 simultaneously-broken mechs in the 30s round budget — a scenario we have never seen in the wild. The parallel path is not worth the framework-integration complexity it introduced (custom-callback dispatch bypassing `get_callback_request`/`try_send`, behaviour staying in `RUNNING` instead of `WAITING_MESSAGE`, late-response cleanup race that can raise `ABCIAppInternalError → process crash` under `stop_and_exit`).

## What this PR does

1. **Restores the sequential `get_http_response` loop** in `populate_tools`. This is the framework-integrated path that worked in v0.25.0.
2. **Adds per-CID coalescing** as #69's proper successor: pending mechs are grouped by `mech.service.metadata_str` and fetched once per distinct CID. The resulting tool set is broadcast to every mech in the group. Omen's 4-identical-CIDs case collapses to **1 fetch** (vs 4 sequential or 4 parallel) — strictly fewer network calls than either prior approach.
3. **Preserves all quarantine + retry semantics** from #66 and #68:
   - Permanent errors (4xx, 5xx with marker) → immediate group-wide quarantine, no retries
   - Transient retries-exhausted → group-wide quarantine after retry budget
   - Empty manifests (deterministic per-CID) → group-wide quarantine
4. **Drops `mech_tools_parallel_timeout`** from `MechParams` and `skill.yaml`.
5. **Replaces** `TestFetchHttpParallel` and `TestPopulateToolsParallelSemantics` with sequential + CID-grouping coverage.

## Test plan

- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests -rfE` — 356 passed.
- [x] `uv run tomte check-code` — all 6 linters green (black, isort, flake8, mypy, pylint, darglint).
- [x] `make generators` — skill hash regenerated; `packages.json` updated.
- [x] **Manual verification of the Omen 4-identical-CIDs case in test** — `TestCidGrouping.test_shared_cid_fetched_once_for_all_mechs` asserts 4 mechs sharing a CID result in exactly 1 HTTP fetch.

## Notes

- This restores semantics that were live-validated in v0.25.0; the trader has already been pinned to v0.25.0 in Pearl, so this brings `main` back in line with what's shipped.
- For the production hang report and root-cause analysis, see `MECH_INFORMATION_HANG_REPORT.md` (local).

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -13,7 +13,7 @@
         "contract/valory/agreement_store_manager/0.1.0": "bafybeifip7e6al6nwtwbyqol4xm2iqnuycrdefdie7vyxlxeh33ubqeooa",
         "contract/valory/transfer_nft_condition/0.1.0": "bafybeie267gridf6z2yxk7cvt5fdy66cocyg5dn3rj7zvnvl3otnk6iw34",
         "contract/valory/subscription_provider/0.1.0": "bafybeibulnegkf7jsz4lrhvsvai65wytjspl6itkhb6dhbl3fwsbpbyh6u",
-        "skill/valory/mech_interact_abci/0.1.0": "bafybeidq4vfhz5hvixmzor4epnnh2xzhhzedcxhxfoa6nqgqjp3yqhktdu"
+        "skill/valory/mech_interact_abci/0.1.0": "bafybeihazofcrox6qvvxojxm4t7ezic4llxwvleoalfk67vqdevukwy3zi"
     },
     "third_party": {
         "protocol/valory/acn_data_share/0.1.0": "bafybeieqisfr6lhaaesq37lk7vyaxza2h6nucshksgyo2djqfyf2saqo4i",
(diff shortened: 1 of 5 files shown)


Adamantios (approved):

bennyjo (approved):
LGTM — clean revert; per-CID coalescing is a strict win for the Omen 4-identical-CIDs case; quarantine + retry semantics preserved.

DhairyaPatel7 (approved):

PR #75, opened 2026-04-30
chore: Wave 2a + 2b — tomte canonical migration

DhairyaPatel7:
## Summary

Centralises tooling state through tomte 0.7.0 and consumes its auto-derived defaults. Mirrors the Wave 2a/2b shape landed in [trader#925](https://github.com/valory-xyz/trader/pull/925) and [mech#437](https://github.com/valory-xyz/mech/pull/437). Also bumps the open-autonomy / open-aea framework pins to the latest releases (0.21.20 / 2.2.3) and addresses a number of typing / dataclass issues that strict mypy surfaced along the way.

## Wave 2a — tomte tox runtime migration

Per-repo identity moves to `[tool.tomte]` in `pyproject.toml`; the ~700-line hand-written `tox.ini` shrinks to a `[tomte-extensions]` + `[mypy-*]` overlay consumed by `tomte tox` at runtime.

**Workflow:** `tox -e <env>` → `tomte tox -e <env>` everywhere; drop spell-check step; `tomte check-copyright --author valory` (third-party packages auto-excluded via `packages.json`); modernize codecov to v5.

**Files removed:** `.pylintrc`, `.spelling` — folded into tomte canonical.

**Files added to `.gitignore`:** `.tomte-tox.ini`, `.darglint`, `.gitleaks-merged.toml` (runtime artifacts from `tomte tox`).

## Wave 2b — canonical pytest defaults

Drop the per-repo `[tool.pytest.ini_options]` block; canonical markers (`integration`, `e2e`) and the `aea.*` `DeprecationWarning` filter now ship via tomte's `pytest11` entry-point plugin. Local `[pytest]` overlay in `tox.ini` keeps repo-specific markers and `log_cli` settings.

## Tomte v0.7.0 + framework bump

- Replace pre-release tomte SHA with the **v0.7.0 tag** in `pyproject.toml`, `[tool.tomte] tomte_dep_pin`, and every CI workflow install step. Inline `# TODO: flip to ==0.7.0` next to each.
- Bump open-autonomy / open-aea-test-autonomy / open-aea-helpers: **0.21.19 → 0.21.20**.
- Bump open-aea / open-aea-* runtime deps: **2.2.1 → 2.2.3**.
- Refresh `packages/packages.json` third_party hashes against `valory-xyz/open-autonomy@v0.21.20` and `valory-xyz/open-aea@v2.2.3` (synced via `autonomy packages sync --update-packages`); re-lock dev hashes.
- Workaround anchorpy 0.20.2 (newly transitive via open-aea-ledger-solana): its pytest11 entry point imports a stale `pytest_xprocess.getrootdir` and crashes pytest collection. Disabled via `addopts = -p no:pytest_anchorpy`. No tests use anchorpy fixtures.
- `[Authorized Packages]` block whitelists anchorpy / based58 / jsonalias (MIT per their GitHub repos, no PyPI License field) and httpx (BSD-3-Clause; uv resolves a 1.0.dev* build under `prerelease = "allow"` whose metadata ships no License field; pinned to `>=1.0.0-dev,<2`).

## Type-fixes / behaviour cleanups (review pass)

- **`MechRequest.data` typing**: was declared `str` but populated as `bytes` from contract events; consumer papered over with `# type: ignore[attr-defined]`. Changed to `Union[str, bytes]`, dropped the type-ignore, added an `isinstance` guard at the comparison site, and taught `DataclassEncoder` to emit hex for bytes so JSON serialisation still works.
- **Six casts that lied to mypy**: replaced `cast(str, self.subscription_address)` / `cast(bytes, self.lock_id)` etc. with explicit None guards at the call sites (`_subscription_contract_interact`, `_build_create_fulfill_tx_data`) so a missing fetch logs and returns `False` rather than passing `None` to `contract_interact(contract_address=None, ...)` or crashing on `None.hex()`.
- **`_extract_and_set_receivers`**: distinguishes "DDO not fetched" from "no service entry in DDO" instead of conflating them via `or {}`.
- **`_build_create_*_tx_data` (3 sites)**: use `self._<x>_tx_data` directly so a single error log fires per failure (was: property-accessor's log followed by the explicit one).
- **`_prepare_safe_tx`**: split the conflated check; treat `mech_max_delivery_rate is None or total_nvm_balance is None` as fetch failure (`return False`) instead of as "go buy a subscription".
- **`models.py` cast/enforce ordering**: `enforce(... is not None)` now precedes the `cast(str, ...)` so the runtime guard reflects the type narrowing.

## Test pass cleanups

- 5 test files: `: MagicMock,` annotations on parametrize params replaced with `: Any,` (the values are `PaymentType` / `str` / `int` / `bool`, not mocks); fixture params replaced with the proper class — `MechRequestBehaviour`, `MechResponseBehaviour`, `MechPurchaseSubscriptionBehaviour`, `ConcreteBehaviour`.
- Test `pytest_targets` widened to include the three contract test packages (`mech`, `mech_mm`, `mech_marketplace_legacy`) — the previous comment that "no Python tests yet" was stale and the previous `tox.ini` ran them. **423 tests now run vs. 361 before.**

## CI fixes uncovered along the way

- **Copyright year:** bump `mech_interact_abci/rounds.py` to `2023-2026`; re-lock packages.
- **`tomte check-doc-links`:** restored as a workflow step.
- **Bytes pass-through (merged from main):** test/contract drift — properties now return `bytes` directly (no string parsing); resolved merge by taking main's bytes-passthrough version of `test_purchase_subscription.py` and `test_request.py`.

## Test plan

All workflow jobs green:
- `test` matrix: 3 OS × 5 Python versions (3.10–3.14) — 423 tests pass.
- `linter_checks`, `lock_check`, `copyright_and_dependencies_check` (liccheck, check-dependencies, check-third-party-hashes, check-doc-links, check-doc-hashes), `gitleaks`.
- `Dry-release framework`, CodeQL.


DhairyaPatel7:
Pushed `0914fc3` to bump tomte / open-aea / open-autonomy in lockstep:

**Tomte SHA → tag**
Replaced the pre-release `tomte.git@54780e1e` SHA with the v0.7.0 tag in `pyproject.toml`, `[tool.tomte] tomte_dep_pin`, and every CI workflow install step.

**Framework bumps**
- open-autonomy / open-aea-test-autonomy / open-aea-helpers: `0.21.19 → 0.21.20`
- open-aea / open-aea-* runtime deps: `2.2.1 → 2.2.3`
- pyproject.toml + 13 dev contract YAMLs + skill/ipfs connection YAML

**Third-party hash refresh**
- `packages/packages.json` third_party hashes refreshed against `valory-xyz/open-autonomy@v0.21.20` and `valory-xyz/open-aea@v2.2.3` via `autonomy packages sync --update-packages`.
- Dev hashes re-locked.

**Workarounds for fleet-wide regressions from the OA bump**
- anchorpy 0.20.2 (newly transitive via open-aea-ledger-solana) ships a pytest11 entry point importing a stale `pytest_xprocess.getrootdir` and crashes pytest collection. Disabled via `addopts = -p no:pytest_anchorpy`. No tests use anchorpy fixtures.
- Extended `[Authorized Packages]` for transitive deps with no PyPI-declared license: anchorpy / based58 / jsonalias (MIT per their GitHub repos) and httpx (BSD-3-Clause; uv resolves a 1.0.dev3 build under `prerelease = "allow"` whose metadata ships no License field).

All local CI green: `lock_check`, `copyright_and_dependencies_check` (liccheck, check-dependencies, check-third-party-hashes, check-doc-hashes), `linter_checks` (safety, bandit, check-packages, check-hash, black-check, isort-check, flake8, mypy, pylint, darglint, check-abci-docstrings, check-abciapp-specs, check-handlers), `gitleaks`, and `py3.10-darwin` tests (361 passed).

LOCKhart07 (commented):
## Review

### Blocking

**1. Contract pytest tests silently excluded from CI** — `pyproject.toml:60-63`

```toml
pytest_targets = [
    "packages/valory/skills/mech_interact_abci/tests",
]
```

The accompanying comment ("the contracts under `packages/valory/contracts/` have no Python tests yet") is factually wrong. Three substantive contract test files exist on `main` (added in `efe94b0` and refined in `fd350fd` / `14e0d54`):

- `packages/valory/contracts/mech/tests/test_contract.py`
- `packages/valory/contracts/mech_mm/tests/test_contract.py`
- `packages/valory/contracts/mech_marketplace_legacy/tests/test_contract.py`

The old `tox.ini` ran all of them with coverage (`-rfE {env:CONTRACTS_PATHS}/mech` etc., plus matching `--cov=` flags). The new config drops them entirely. The PR even bumps their fingerprints and adds type annotations in the diff, so the regression is invisible from the diff alone — they still get linted, just never executed. The "361 tests pass" claim therefore represents a coverage regression, not parity with `main`.

Fix: add the contract test paths back to `pytest_targets` (and at minimum match the old set: `agent_registry`, `agent_mech`, `erc20`, `mech`, `mech_mm`, `mech_marketplace_legacy`), and correct the comment.

### Non-blocking

**2. `cast` before `enforce` is misleading** — `models.py:290-294`

```python
self.agent_registry_address: str = cast(str, kwargs.get("agent_registry_address"))
enforce(self.agent_registry_address is not None, "Agent registry address not specified!")
```

The cast asserts `str` to the type checker before the runtime guard runs. A future reader can trust the `: str` annotation and remove the `enforce`. The `multisend_address` block a few lines above does it correctly (local var → enforce → assign). Either match that pattern or just drop the cast.

**3. `Check doc links` step silently removed** — `.github/workflows/main_workflow.yml`

The previous workflow ran `tomte check-doc-links` with a curated `--url-skips` list. The new `copyright_and_dependencies_check` job drops it. The PR description mentions dropping spell-check but not doc-links. If intentional (e.g. tomte 0.7 dropped the command), please mention it; otherwise restore.

**4. `# type: ignore[attr-defined]` on `request.data.hex()`** — `behaviours/response.py:520`

This papers over a real type bug (`MechRequest.data: str` but populated as `bytes` at runtime from the contract event payload). The commit message acknowledges fixing the dataclass is out of scope — fine — but please file a follow-up issue so the `type: ignore` doesn't rot indefinitely.

**5. `httpx: >=0` in `[Authorized Packages]`** — `tox.ini`

Broadest possible pin; won't catch a future incompatible httpx 1.x. Either narrow to `<2`, or pin to the `1.0.dev*` series that's actually resolving under `prerelease = "allow"`.

DIvyaNautiyal07:
## Review

### P1 (Should Fix)

- **Test type annotations are factually wrong.** In `tests/behaviours/test_request.py` (around lines 114-119, 134-138, 152-156) and sibling files, parametrize params are annotated `MagicMock` but the runtime values are `PaymentType.NATIVE` / `str` / `int` / `bool`; fixture params annotated `MagicMock` are actually `MechRequestBehaviour` / `MechResponseBehaviour` / `AcnHandler` instances. `MagicMock` satisfies mypy because every attribute access is permissive, so the strict-mypy gate accepts a lie. Fix: use `Any` for parametrize params, the proper class type for fixtures (the conftest already does this — `fixture() -> MechResponseBehaviour`).

- **Six casts that hide invariants** rather than enforce them:
  - `behaviours/request.py:474` `cast(str, self.subscription_address)`
  - `behaviours/purchase_subcription.py:330` `cast(str, self.from_address)`
  - `behaviours/purchase_subcription.py:336, 362` `cast(bytes, self.lock_id).hex()`
  - `behaviours/purchase_subcription.py:354` `cast(List[str], self.receivers)`
  - `behaviours/purchase_subcription.py:364` `cast(bytes, self.transfer_id).hex()`

  Each silences the `Optional[...]` warning. If the upstream fetch failed and the underlying `_*` attribute is `None`, the cast lies and either passes `None` to `contract_interact(contract_address=None, ...)` or crashes on `None.hex()`. The PR already adds explicit `None` guards for `mech_max_delivery_rate`, `total_nvm_balance`, and `ddo_endpoint` — be consistent and add guards (`if x is None: return False`) at these six sites instead of casting.

### P2 (Consider)

- **`behaviours/response.py:316`** uses `# type: ignore[arg-type]` (specific code) while every other `Performative.GET_RAW_TRANSACTION` site (10+ in this skill) uses bare `# type: ignore`. Inconsistent — pick one.
- **`behaviours/response.py:519-523`** `request.data.hex()  # type: ignore[attr-defined]`: the comment correctly identifies the dataclass mistype (`MechRequest.data: str` declared but populated with `bytes`). The real fix is at `states/base.py:90` — change to `data: bytes = b""` or `Union[str, bytes]`. Worth a follow-up issue.
- **`behaviours/request.py:929-934`**: new `or` chain returns `True` (= "buy subscription") when `mech_max_delivery_rate` or `total_nvm_balance` is `None`. Previously would have been a `TypeError`. Treating "fetch failed" as "go buy a subscription" is a behavior change. Practically guarded today by `_get_price` validating earlier, but worth noting.
- **`behaviours/purchase_subcription.py:386-410`** `_extract_and_set_receivers`: previously crashed on `ddo_values=None`; now silently logs "No service found in DDO." Behavior change from crash → silent skip.
- **`behaviours/purchase_subcription.py:610, 642, 673`**: `if not status or self.agreement_tx_data is None` triggers two back-to-back error logs (property accessor logs first, then the explicit log). Use `self._agreement_tx_data` for the guard.

### Nit

- **`states/base.py:282`** `is_dataclass(obj) and not isinstance(obj, type)` is a real correctness fix (encoding a class object would crash on `asdict`). Good.
- **`rounds.py:171`** `Set[str]` → `FrozenSet[str]` matches the parent type — correct fix.
- **`models.py:263, 290`** `cast(str, ...)` is written *before* `enforce(... is not None ...)`. Reads backwards but works at runtime.

DhairyaPatel7:
Pushed `849a071` addressing review:

**@LOCKhart07 — Blocking**
- pyproject.toml: added `mech / mech_mm / mech_marketplace_legacy` contract test paths back to `pytest_targets`. The comment claiming no contract tests existed was stale — they shipped on main and the previous `tox.ini` ran them with coverage. **423 tests now run vs. 361 previously.**

**@LOCKhart07 — Non-blocking**
- main_workflow.yml: restored `tomte check-doc-links` step.
- tox.ini: narrowed the httpx `[Authorized Packages]` entry from `>=0` to `>=1.0.0-dev,<2`. semantic_version (used by liccheck) rejects PEP 440 dev specifiers, so the SemVer form is the closest to the resolved 1.0.0-devN tag.
- models.py:290-294: reordered `cast()` to follow `enforce(... is not None ...)`, matching the `multisend_address` pattern at L261-263.
- Added TODOs at both `git+https` tomte references in pyproject.toml.

**@DIvyaNautiyal07 — P1 (six casts that hide invariants)**
- Dropped the four `cast(...)` lies on `Optional[...]` accessors in `behaviours/request.py` (`subscription_address`) and `behaviours/purchase_subcription.py` (`from_address` / `lock_id` / `transfer_id` / `receivers`).
- Guarded at the call sites (`_subscription_contract_interact`, `_build_create_fulfill_tx_data`) so a missing fetch logs and returns False instead of silently passing `None` to `contract_interact(contract_address=None, ...)` or crashing on `None.hex()`.

**Deferred to follow-up issues** (filed separately):
- Test type annotations using `MagicMock` for parametrize params that are actually `PaymentType` / `str` / `int` / `bool` — needs sweep across 4 test files; tracked separately.
- `# type: ignore[attr-defined]` papering over `MechRequest.data: str` declared but populated as `bytes`. Real fix is at `states/base.py:90`; tracked separately.
- `purchase_subcription.py` behaviour-change concerns (or-chain, `_extract_and_set_receivers`, back-to-back error logs) — pre-existing semantics; tracked separately.

All local CI green: `lock_check`, `copyright_and_dependencies_check` (liccheck, check-dependencies, check-third-party-hashes, check-doc-links, check-doc-hashes), `linter_checks`, `gitleaks`, `py3.10-darwin` (423 tests).

DhairyaPatel7:
Pushed `0a4ac22` pulling in everything I'd previously deferred to follow-ups (per @DIvyaNautiyal07's full review):

**Test type annotations (P1, 4 files)**
Replaced `: MagicMock,` annotations with the actual types:
- Parametrize parameters → `: Any,` (the values are `PaymentType` / `str` / `int` / `bool`, not mocks).
- Fixture parameters → the proper class — `MechRequestBehaviour`, `MechResponseBehaviour`, `MechPurchaseSubscriptionBehaviour`, `ConcreteBehaviour`. The conftest fixtures already declare these return types, so the call sites now match.

Strict mypy then surfaced ~11 pre-existing `behaviour.contract_interact = mock_fn` style monkey-patches that weren't visible while the receiver was typed as `MagicMock`. Those are tagged with `# type: ignore[method-assign,assignment]` since they're a deliberate test-only pattern.

**`MechRequest.data` type fix (P2 follow-up)**
- `states/base.py:90`: changed `data: str = ""` → `data: Union[str, bytes] = ""` reflecting the dual usage (bytes from contract event payloads, hex string when carried through synchronized data).
- `behaviours/response.py:519-523`: dropped the `# type: ignore[attr-defined]`, added an `isinstance(request.data, bytes)` guard so the `.hex()` call only fires when needed.
- `utils.DataclassEncoder.default`: encode bytes as hex string so JSON-serialising a MechRequest with a real bytes payload no longer crashes.
- Test sites in `test_response_behaviour.py` switched from `MagicMock().hex.return_value = "..."` to real bytes (`b"\xaa\xbb\xcc"`) so the new isinstance branch is exercised.

**`purchase_subcription.py` behaviour-change concerns (P2)**
- `_extract_and_set_receivers`: distinguish "DDO not fetched" from "no service entry in DDO" instead of conflating them via `or {}`.
- `_build_create_*_tx_data` (3 sites): use `self._<x>_tx_data` directly in both the None guard and the subsequent `MultisendBatch(data=…)` call so a single error log fires on failure instead of the property-accessor's log followed by the explicit one.

**`request.py` or-chain semantics (P2)**
- `_prepare_safe_tx`: split the conflated check. `mech_max_delivery_rate is None or total_nvm_balance is None` is now a fetch failure (log error, return False); the legitimate "balance < max_rate" case still returns True for "buy a subscription".

All 423 unit tests + lint suite + check-hash / check-dependencies / check-third-party-hashes / liccheck / gitleaks pass locally.

DIvyaNautiyal07 (dismissed):
LGTM — most review items addressed:
- Test annotations fixed (`MagicMock` → `Any` / proper class types)
- `request.py:474` cast replaced with explicit None guard
- `MechRequest.data` is now `Union[str, bytes]`; `response.py:521` does `isinstance` dispatch — no more `# type: ignore[attr-defined]`
- `request.py:929` adds explicit None guard with logged error
- `purchase_subcription.py` `_extract_and_set_receivers` and `_agreement_tx_data` guards both fixed

Two follow-up items, neither blocking:

1. Five `cast(...)` calls remain in `purchase_subcription.py:330, 336, 354, 362, 364` inside the `fulfill_params` / `fulfill_for_delegate_params` properties (both `# pragma: no cover`). These are defensible if the orchestrator guarantees `from_address` / `lock_id` / `transfer_id` / `receivers` are populated before the property is read, but the original concern (cast lies if any upstream fetch failed silently) stands. Worth either an explicit precondition assert at the top of each property, or a quick comment explaining why the casts are safe here vs. the request.py site that got a None guard.

2. `response.py:316` still uses `# type: ignore[arg-type]` while every other `Performative.GET_RAW_TRANSACTION` site uses bare `# type: ignore`. Cosmetic, can land in a follow-up.

Approving.

packages/valory/skills/mech_interact_abci/behaviours/purchase_subcription.py
```diff
@@ -652,27 +683,29 @@ def _build_create_fulfill_tx_data(
             fulfill_params=self.fulfill_params,
             chain_id=self.params.mech_chain_id,
         )
```
OjusWiZard, line 685: **HIGH** — This new None-guard correctly prevents `None.hex()` by returning `False`, but the caller in `_prepare_safe_tx` (`yield from self.wait_for_condition_with_sleep(step)`) never inspects the return value. The `False` is silently discarded and the loop continues into `_build_multisend_data` / `_build_multisend_safe_tx_hash` with incomplete state. Compare to `request.py` `_prepare_safe_tx` which checks `status` at each step. Until the loop respects the bool, the safety guarantee here is illusory.

packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -582,7 +593,7 @@ def setup(self) -> None:  # pragma: no cover
         self._mech_requests = self.synchronized_data.mech_requests
         self.context.logger.info(f"Processing mech requests: {self._mech_requests}")
         self.context.logger.info("Selecting priority mech...")
-        self.priority_mech_address = self.get_priority_mech_address()
+        self.priority_mech_address = self.get_priority_mech_address() or ""
```
OjusWiZard, line 596: **MEDIUM** — `get_priority_mech_address() or ""` masks `None` with an empty string. The next log line then prints `'' was selected as priority mech.` and downstream contract calls fail with an opaque "invalid address" rather than an actionable upfront error. This is inconsistent with the explicit-None-guard pattern this PR otherwise establishes — prefer detecting `None`, logging, and short-circuiting.

packages/valory/skills/mech_interact_abci/behaviours/purchase_subcription.py
```diff
@@ -139,7 +150,10 @@ def ddo_register(self) -> Optional[List]:
     def ddo_endpoint(self) -> Optional[str]:
         """Return the ddo endpoint."""
         try:
-            return self.ddo_register[DDO_ENDPOINT_IDX]
+            ddo_register = self.ddo_register
+            if ddo_register is None:
```
OjusWiZard, line 154: **MEDIUM** — Asymmetric error handling: the `IndexError` branch below logs, but this `ddo_register is None` branch returns `None` silently. `_get_ddo_data` happens to log on its own, but other current/future callers reading the property for diagnostics lose the distinction between "register unfetched" and "DDO has no endpoint entry". Add a `logger.error` here or document the contract.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -84,10 +84,13 @@ def populate_tools(
...
         """
         pending_by_cid: Dict[str, List[Any]] = {}
-        for mech in mech_info:
+        for mech in mech_info or []:
             if mech.relevant_tools or mech.address in self._failed_mechs:
                 continue
-            pending_by_cid.setdefault(mech.service.metadata_str, []).append(mech)
+            metadata_str = mech.service.metadata_str
+            if metadata_str is None:
+                continue
```
OjusWiZard, line 92: **MEDIUM** — Silent skip: mechs without `metadata_str` are dropped from `populate_tools` with no log, leaving operators no signal why a mech never gets tools populated. Add `self.context.logger.warning(f"Skipping mech {mech.address}: missing service metadata_str")` before `continue`.

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -302,7 +307,7 @@ def mechs_info(self) -> MechsInfo:
         mechs_info = self.db.get("mechs_info", SERIALIZED_EMPTY_LIST)
         if isinstance(mechs_info, str):
             mechs_info = json.loads(mechs_info)
-        return [MechInfo(**item) for item in mechs_info]
+        return [MechInfo(**item) for item in (mechs_info or [])]
```
OjusWiZard, line 310: **MEDIUM** — `or []` collapses `None` and any falsy non-list (`0`, `""`, `{}`) into "no work to do," indistinguishable from a corrupted/missing key. A misconfigured round transition that fails to populate `mechs_info` will look identical to steady state. Distinguish `None` (uninitialized → empty OK, debug-log) from unexpected types (error-log) before defaulting. Same pattern at lines 381 (`mech_requests`) and 391 (`mech_responses`).

packages/valory/skills/mech_interact_abci/utils.py
```diff
@@ -29,12 +29,18 @@ class DataclassEncoder(json.JSONEncoder):
 
     def default(self, o: Any) -> Any:
         """The default JSON encoder."""
+        if isinstance(o, bytes):
```
OjusWiZard, line 32: **Test gap** — Both new bytes paths in `DataclassEncoder` (this top-level branch and the `asdict`-walk loop below) lack tests in `tests/test_utils.py`. A regression that drops either branch silently reintroduces `TypeError: bytes is not JSON serializable`. Add `test_encode_top_level_bytes` (e.g. `b"\xab"` → `"ab"`) and `test_encode_dataclass_with_bytes_field` (MechRequest with `data=b"\xab"`).

packages/valory/skills/mech_interact_abci/tests/behaviours/test_response.py
```diff
@@ -21,6 +21,9 @@
 
 from unittest.mock import MagicMock
 
+from packages.valory.skills.mech_interact_abci.behaviours.response import (
```
OjusWiZard, line 24: **Test gap** — `MechRequest.data` is now `Union[str, bytes]`, but `_is_legacy_match` is not exercised on both arms (a `bytes` value that must `.hex()`, and a `str` value that must pass through). This is the central type-fix of the PR — please add a parametrized test covering both.

OjusWiZard (dismissed):
Approving with inline comments. Infra migration is solid; type-fixes are directionally right. The HIGH item below is worth addressing before merge — without it the new None-guard is cosmetic. The other comments are non-blocking but consistent with this PR's cleanup spirit.

**Verified non-issues:**
- `_extract_and_set_receivers` DDO-not-fetched branch logs distinctly. ✅
- `_build_create_*_tx_data` consolidation drops no log. ✅
- `request.py` `_prepare_safe_tx` fetch-failure split logs before returning False. ✅
- `DataclassEncoder` bytes ordering is correct. ✅
- `: MagicMock,` → `: Any,` parametrize swap is cosmetic. ✅
- 423 vs 361 test count holds up; contract test packages are genuine new coverage.

DhairyaPatel7:
Pushed `e1c0580` — `tomte check-copyright` flagged `utils.py` because the file picked up edits in 2026 but the header still read `2023-2025`. Bumped to `2023-2026`. CI should now go green.

DIvyaNautiyal07 (approved):

OjusWiZard (approved):

PR #76, opened 2026-04-30
fix(request): pass approval bytes to MultisendBatch unchanged

LOCKhart07:
## Summary

The marketplace-v2 token-payment request path crashed on the very first request:

```python
data=bytes.fromhex(self.approval_data.removeprefix("0x").removeprefix("0X"))
```

`self._approval_data` is `bytes` (returned by `ERC20TokenContract.build_approval_tx` as `bytes.fromhex(data[2:])`, preserved through `DictProtobufStructSerializer`), so calling `.removeprefix("0x")` on it raises `TypeError: a bytes-like object is required, not 'str'`. The bug was masked from mypy by an incorrect `Optional[str]` annotation that has been wrong since the feature shipped.

The path is gated on marketplace-v2 + `using_token` (USDC/OLAS direct ERC20), which is why it didn't surface in native or NVM flows. Both `_approve_balance_tracker` and `_build_token_approval` carried `# pragma: no cover`, so CI never exercised it either.

## Regression history

The original feature shipped correctly:

- `57d078b` (2025-10-10, "feat: approve the balance tracker for spending tokens") — used `data=HexBytes(self.approval_data)`. ✅ Worked: `HexBytes` accepts `bytes`.
- `8b2e5b2` (2026-04-23, "chore(skill): drop hexbytes + web3 from mech_interact_abci") — replaced `HexBytes(self.approval_data)` with `bytes.fromhex(self.approval_data.removeprefix("0x"))`. ❌ **Bug introduced here.** The dep-cleanup refactor trusted the (wrong) `Optional[str]` annotation and rewrote the line as if `approval_data` were a hex string.
- `82d2867` (2026-04-24, "chore(review): address Divya feedback") — added a second `.removeprefix("0X")` for case-insensitivity. Cosmetic; bug already present.

So the broken state has only existed for ~7 days, and only on the marketplace-v2 token-payment path. The lying type annotation is what let the refactor land.

## Changes

- `behaviours/request.py`:
  - Annotation fix: `_approval_data` and the `approval_data` property are now `Optional[bytes]`.
  - Pass `self.approval_data` straight to `MultisendBatch.data` (already accepts `bytes`/`bytearray`); drop the broken `bytes.fromhex(...removeprefix...)` dance.
  - Remove `# pragma: no cover` from `_approve_balance_tracker` and `_build_token_approval`.
- `tests/behaviours/test_request.py`:
  - `TestApproveBalanceTracker` (3 tests) — asserts every kwarg passed to `contract_interact` (callable, `data_key`, `placeholder`, spender, amount, chain_id), failure propagation, and the info log.
  - `TestBuildTokenApproval` (5 tests) — including a regression guard that asserts `MultisendBatch.data == b"\xab\xcd\xef"` and `isinstance(batch.data, bytes)`. If anyone re-introduces the `bytes.fromhex(...removeprefix...)` line, this test fails with `TypeError` before reaching the assertion.

Coverage on `behaviours/request.py` is 99% (one pre-existing miss at `select_mech` line 571, unrelated). The two methods touched here are 100% covered.

## Test plan

- [x] `uv run tomte check-code` — black, isort, flake8, mypy, pylint, darglint all clean.
- [x] `uv run tomte check-security` — safety + bandit clean.
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/behaviours/test_request.py` — 47/47 pass.
- [x] `uv run make generators` — package hashes locked.
- [ ] Smoke test on a marketplace-v2 mech configured for `TOKEN_USDC` or `TOKEN_OLAS`: confirm the approval batch builds and the multisend lands on-chain.

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/request.py
@@ -114,7 +114,7 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
         self._subscription_address: Optional[str] = None
         self._subscription_id: Optional[int] = None
         self._balance_tracker: Optional[str] = None
-        self._approval_data: Optional[str] = None
+        self._approval_data: Optional[bytes] = None
 
     @property
     def metadata_filepath(self) -> str:
@@ -253,7 +253,7 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
         return self._balance_tracker
 
     @property
-    def approval_data(self) -> Optional[str]:
+    def approval_data(self) -> Optional[bytes]:
         """Get the approval data."""
         if self._approval_data is None:
             self.context.logger.warning(
@@ -743,7 +743,7 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
             )
         )
 
-    def _approve_balance_tracker(self) -> WaitableConditionType:  # pragma: no cover
+    def _approve_balance_tracker(self) -> WaitableConditionType:
         """Build approval for the balance tracker."""
         self.context.logger.info("Building approval for token payment.")
 
@@ -761,7 +761,7 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
             )
         )
 
-    def _build_token_approval(self) -> WaitableConditionType:  # pragma: no cover
+    def _build_token_approval(self) -> WaitableConditionType:
         """Get the balance tracker, build approval for the token payment and add it to the multisend batch."""
         if not self._balance_tracker:
             status = yield from self._get_balance_tracker()
@@ -776,9 +776,7 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
 
         batch = MultisendBatch(
             to=self.params.price_token,
-            data=bytes.fromhex(
-                self.approval_data.removeprefix("0x").removeprefix("0X")
-            ),
+            data=self.approval_data,
         )
         self.multisend_batches.append(batch)
         self.context.logger.info("Successfully built approval data.")
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -13,7 +13,7 @@
         "contract/valory/agreement_store_manager/0.1.0": "bafybeifip7e6al6nwtwbyqol4xm2iqnuycrdefdie7vyxlxeh33ubqeooa",
         "contract/valory/transfer_nft_condition/0.1.0": "bafybeie267gridf6z2yxk7cvt5fdy66cocyg5dn3rj7zvnvl3otnk6iw34",
         "contract/valory/subscription_provider/0.1.0": "bafybeibulnegkf7jsz4lrhvsvai65wytjspl6itkhb6dhbl3fwsbpbyh6u",
-        "skill/valory/mech_interact_abci/0.1.0": "bafybeiaczcber6tmw4rrlbislsksahpj4wmnhyeg5tgxumkxhm3dzjxhqu"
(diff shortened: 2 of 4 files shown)


packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -776,9 +776,7 @@ def _build_token_approval(self) -> WaitableConditionType:  # pragma: no cover
 
         batch = MultisendBatch(
             to=self.params.price_token,
-            data=bytes.fromhex(
-                self.approval_data.removeprefix("0x").removeprefix("0X")
-            ),
+            data=self.approval_data,
```
bennyjo, line 779: The same regression introduced by `8b2e5b2` looks like it's still present in `behaviours/purchase_subcription.py` for three sibling properties:

- `_subscription_token_approval_tx_data` (line 121) — filled by the *same* `build_approval_tx` callable used here (line 625)
- `_agreement_tx_data` (line 120) — filled by `build_create_agreement_tx` (`nft_sales/contract.py:74`, also returns `bytes.fromhex(data[2:])`)
- `_fulfill_tx_data` (line 122)

All three are typed `Optional[str]` and then decoded via `bytes.fromhex(self._foo.removeprefix("0x").removeprefix("0X"))` (lines 287-288 / 299-302 / 313-315). Since `BaseBehaviour.contract_interact` (`base.py:206`) does `setattr(self, placeholder, data)` with whatever the contract returned, those placeholders should land as bytes too — same `TypeError` the marketplace-v2 path was hitting. The `using_token` gating explains why it hasn't crashed yet, not why it wouldn't.

Fold into this PR or follow-up?

bennyjo (commented):
Scope question — same bytes-vs-str pattern appears in purchase_subcription.py.

packages/valory/skills/mech_interact_abci/behaviours/request.py
LOCKhart07: Yes. Just audited the codebase. Same issue exists at three locations. Would be a separate PR so as to not increase scope of this one

bennyjo (approved):
Approving — fix is correct and well-tested for its scope. Confirmed `_approval_data` is bytes (via `ERC20TokenContract.build_approval_tx` → `BaseBehaviour.contract_interact` `setattr`), and `MultisendBatch.data` accepts bytes/bytearray. Parallel bug in `purchase_subcription.py` (3 properties from the same `8b2e5b2` commit) tracked as a separate PR per author.

packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -776,9 +776,7 @@ def _build_token_approval(self) -> WaitableConditionType:  # pragma: no cover
...
         )
```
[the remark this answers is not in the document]
LOCKhart07: Skipping the guard — the scenario it protects against can't happen. `contract_interact` (`base.py:200-204`) already returns `False` if the contract response body lacks the `data` key, and only sets the placeholder when `data is not None`. So `_approve_balance_tracker()` returning `True` is a hard guarantee that `_approval_data` is bytes, not `None`.

The `Optional[bytes]` annotation just reflects the `__init__` state and the property's "accessed before built" warning; once the upstream call succeeds, the runtime invariant is non-`None`. Adding a downstream guard would be defending against an impossible state — the kind of dead branch the project style guide pushes back on.

For the "or any non-bytes-like value" case: that would mean the contract method itself returned something other than bytes, which is a contract-layer bug. `MultisendBatch.__post_init__` would raise `ValueError` immediately with a clear message — exactly the failure mode we want, not a silently-swallowed `return False`.

OjusWiZard (approved):

DhairyaPatel7 (approved):

PR #78, opened 2026-05-12
feat(mech_interact_abci): valid_mech_tools allowlist

DIvyaNautiyal07:
## Summary

Implements the curated allowlist described in [VALID_MECH_TOOL_SELECTION_FLOW.md](https://github.com/valory-xyz/launch-predict/blob/codex/discoverability-spec/mech-discoverability/VALID_MECH_TOOL_SELECTION_FLOW.md), simplified per review.

The spec originally called for a per-mech `Dict[address, List[tool]]` allowlist. Review feedback (see [thread](https://github.com/valory-xyz/mech-interact/pull/78#issuecomment-4430451893) on the original commits) showed the per-mech indirection doesn't carry weight: "different mechs offer different tools" is already encoded in each mech's IPFS manifest, and the spec's stated rationale (`0xFakeMech exposing the same tool_X is not valid unless explicitly listed`) is covered by a mech-level allowlist on its own. The PR now ships two flat allowlists.

- New `valid_mechs: List[str]` skill param — flat allowlist of mech addresses, lowercased on load. Replaces the old `ignored_mechs` blocklist in the marketplace subgraph query (`address_in` over the allowlist, not `address_not_in` over a blocklist). Empty default; an empty allowlist with marketplace v2 dynamic selection enabled logs a startup warning.
- New `valid_tools: List[str]` skill param — flat allowlist of tool names, also lowercased on load. `populate_tools` lowercases each mech's IPFS manifest entries before the intersect, so `Prediction-Online` in `service.yaml` matches a `prediction-online` manifest entry. Empty `valid_tools` with v2 dynamic selection logs the same startup warning as `valid_mechs`.
- Static `priority_mech_address` is rejected at runtime when `valid_mechs` is configured (non-empty) and the priority is not in it. Empty `valid_mechs` is treated as "not yet configured" so the bump doesn't regress legacy deployers.
- New `SharedState.last_failure_reason` diagnostic with five codes: `subgraph_unavailable`, `allowlist_not_configured` (empty allowlist), `valid_mech_list_empty` (allowlist configured but subgraph returned no matches), `no_overlap_with_valid_tools` (mechs returned but their manifests don't intersect `valid_tools`), and `no_overlap_with_selected_mechs` (ChatUI-pinned mechs don't serve the selected tool). Consumed by the trader-side ChatUI handler in the follow-up branch.
- New `SynchronizedData.selected_mechs` synced-data property: when non-empty, `relevant_mechs_info` is further restricted to that pinned subset. Trader writes this from its ChatUI pin via tool-selection consensus. Property tolerates malformed db payloads by falling back to `[]` rather than raising every round.
- `ignored_mechs` param removed. Anything not in `valid_mechs` is implicitly excluded.

### Heads up: latent FSM hazard fixed

The old `'", "'.join(self.params.ignored_mechs)` was iterating a `FrozenSet`, whose order depends on `PYTHONHASHSEED`. Different agents in the same round could produce different POST bodies to the subgraph and disagree on the `mechs_info` payload, breaking consensus. The new code path uses `sorted(...)` on `valid_mechs` so the query string is deterministic. Called out separately because it's a real fix masked by the param rename.

## Test plan

- [x] Unit tests: 381 passing (`pytest packages/valory/skills/mech_interact_abci/tests/`)
- [x] Static checks clean on touched files: mypy, flake8, pylint, darglint, black, isort
- [x] Live subgraph round-trip against `marketplace-gnosis` confirms `address_in` is accepted and filters exactly to the requested addresses
- [x] `autonomy packages lock --check` passes; `autonomy push-all` published the new skill hash to IPFS
- [ ] Staging smoke test with Pearl team: deploy trader with a 2-mech allowlist, observe one successful round and one deliberate failure (allowlist empty → `allowlist_not_configured` in logs; pinned mech doesn't serve picked tool → `no_overlap_with_selected_mechs`)

## Follow-ups (not in this PR)

- Trader-side changes (`decision_maker_abci`, `chatui_abci`): bump the mech-interact hash, expose `selected_mechs` in chat config, intersect tool selection with `valid_tools`, read `last_failure_reason` for user-facing diagnostics. Branch already prepared, will rebase on this once landed.
- Optional: auto-penalize on mech timeout / malformed response. Spec calls out this gap; out of scope here.

## Breaking changes

Documented in `UPGRADING.md` under `Unreleased`. Summary: `ignored_mechs` is gone, replaced by `valid_mechs` + `valid_tools`. Any overlay (`aea-config.yaml`, `service.yaml`) that still sets `ignored_mechs` is silently ignored. Static `priority_mech_address` must be in `valid_mechs` (when configured) or it's runtime-rejected.


packages/valory/skills/mech_interact_abci/skill.yaml
```diff
@@ -223,7 +223,7 @@ models:
       - stabilityai-stable-diffusion-512-v2-1
       - stabilityai-stable-diffusion-768-v2-1
       deliveries_lookback_days: 30
-      ignored_mechs: []
+      valid_mech_tools: {}
```
LOCKhart07: `UPGRADING.md` is not updated for this breaking change. The current file still lists `ignored_mechs` as a new feature in the `v0.22.2 → v0.22.3` section, and this PR removes that param and replaces it with `valid_mech_tools`. Downstream services that overlay `ignored_mechs` in their `aea-config.yaml` / `service.yaml` will silently lose the override. Please add a `Breaking Changes` entry documenting (a) the removal of `ignored_mechs`, (b) the new `valid_mech_tools` param and its semantics, and (c) that an empty allowlist disables mech fetching entirely.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -297,8 +297,10 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:  # pragma: no cover
             "use_acn_for_delivers", kwargs, bool
         )
         self.irrelevant_tools: set = set(self._ensure("irrelevant_tools", kwargs, list))
-        self.ignored_mechs: FrozenSet[str] = frozenset(
-            self._ensure("ignored_mechs", kwargs, List[str])
+        self.valid_mech_tools: Dict[str, FrozenSet[str]] = (
+            self._normalize_valid_mech_tools(
+                self._ensure("valid_mech_tools", kwargs, dict)
```
LOCKhart07: Empty `valid_mech_tools` silently disables all mech fetching. The `skill.yaml` default is `{}`, and `validate_configuration()` doesn’t flag it. A deployer who forgets to override the allowlist will get `valid_mech_list_empty` every round with no startup-time signal that the skill is misconfigured.

Consider either failing validation when `use_mech_marketplace and mech_marketplace_config.use_dynamic_mech_selection and not valid_mech_tools`, or at minimum emitting a loud `logger.warning` from `__init__` so the misconfig shows up in logs immediately rather than after the first failed round.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -146,13 +149,15 @@ def get_mechs_info(
...
+        self.shared_state.last_failure_reason = None
         mech_info = yield from self.fetch_mechs_info()
 
         if self._fetch_status != FetchStatus.SUCCESS:
+            self.shared_state.last_failure_reason = "subgraph_unavailable"
             return None
 
         if not mech_info:
-            # if the info is None or empty, return None
+            self.shared_state.last_failure_reason = "valid_mech_list_empty"
```
LOCKhart07: `valid_mech_list_empty` overloads two distinct failure modes:

1. The deployer didn’t configure an allowlist — `fetch_mechs_info` short-circuits with `[]` before touching the subgraph.
2. The subgraph returned no mechs whose addresses are in the allowlist.

The PR description scopes a follow-up that surfaces these reasons in the trader ChatUI — from a user’s perspective these two cases need very different remediation. Suggest splitting into e.g. `allowlist_not_configured` (when `not self.params.valid_mech_addresses`) vs the current `valid_mech_list_empty` (subgraph returned nothing matching).

packages/valory/skills/mech_interact_abci/tests/test_models.py
```diff
@@ -100,6 +101,52 @@ def test_with_priority_mech(self) -> None:
...
+            MechParams._normalize_valid_mech_tools({123: ["tool_a"]})  # type: ignore[dict-item]
+
+    def test_non_list_value_raises(self) -> None:
+        """A non-list value is a configuration bug."""
+        with pytest.raises(ValueError, match="must be a list"):
+            MechParams._normalize_valid_mech_tools({"0xmech": "tool_a"})  # type: ignore[dict-item]
+
+    def test_non_string_tool_raises(self) -> None:
+        """A non-string tool entry is a configuration bug."""
+        with pytest.raises(ValueError, match="must be a list"):
```
LOCKhart07: `match="must be a list"` is too loose: the validator emits the same error string for *both* the non-list value and the non-string element branches (`not isinstance(tools, list) or not all(isinstance(...))`). This test would still pass if the `not all(...)` check were removed, defeating the intent of the test name.

Tighten to `match=r"must be a list of tool name strings"`, or additionally assert that the offending list (`['tool_a', 1]`) appears in `str(excinfo.value)`.

packages/valory/skills/mech_interact_abci/tests/test_request_behaviour.py
```diff
@@ -311,3 +313,46 @@ def test_marketplace_v2_no_dynamic_returns_config(self, _mock: MagicMock) -> Non
...
+    def test_static_priority_mech_outside_allowlist_returns_none(
+        self, _mock: MagicMock
+    ) -> None:
+        """A configured priority mech missing from `valid_mech_tools` is rejected."""
+        behaviour = _make_request_behaviour()
+        behaviour._context.params.use_mech_marketplace = True
+        behaviour._context.params.valid_mech_tools = {
+            "0xallowed": frozenset({"tool_a"})
+        }
+        behaviour._context.params.valid_mech_addresses = frozenset({"0xallowed"})
```
LOCKhart07: These tests set both `valid_mech_tools` *and* `valid_mech_addresses` directly on the MagicMock (also at L349), but in production `valid_mech_addresses` is a `@property` derived from `valid_mech_tools.keys()` (`models.py:359-362`). The dual-write means the test passes even if the property’s derivation logic later diverges from the manually-set value.

Suggest setting `valid_mech_tools` only and letting the property compute, e.g. via `PropertyMock` on the params spec, so the test exercises the same code path as production.

packages/valory/skills/mech_interact_abci/skill.yaml
DIvyaNautiyal07: Done in 62fc247. UPGRADING.md now has an `Unreleased` section flagging (a) `ignored_mechs` removal, (b) the new `valid_mechs` and `valid_tools` params (after Benny's flat-list refactor on the same commit), and (c) the empty-allowlist semantics + `allowlist_not_configured` reason.

packages/valory/skills/mech_interact_abci/models.py
DIvyaNautiyal07: Done in 62fc247: `validate_configuration` now logs a warning at startup when marketplace v2 dynamic selection is enabled and `valid_mechs` is empty, so a missing override shows up at boot rather than after the first failed round. Couldn't make this a hard validation error because the empty default is the legitimate state for downstream services that haven't been migrated yet.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
DIvyaNautiyal07: Split in 62fc247. `get_mechs_info` now writes `allowlist_not_configured` when `self.params.valid_mechs` is empty (operator never configured one) and keeps `valid_mech_list_empty` for the other case (allowlist configured but the subgraph returned nothing matching). Trader-side ChatUI gets two distinct reasons to surface.

packages/valory/skills/mech_interact_abci/tests/test_models.py
DIvyaNautiyal07: Resolved by the flat-list refactor in 62fc247. `_normalize_valid_mech_tools` is gone (the helper, the dict shape, the 25-line validator, and `TestNormalizeValidMechTools` entirely). `valid_mechs` is now a `List[str]` lowercased into a frozenset at load time, and `valid_tools` is the same shape — both rely on AEA's `_ensure(..., List[str])` for shape validation. No regex test to tighten anymore.

packages/valory/skills/mech_interact_abci/tests/test_request_behaviour.py
DIvyaNautiyal07: Resolved by the refactor in 62fc247. `valid_mech_addresses` was a property derived from `valid_mech_tools.keys()`; both are gone now. `valid_mechs` is the source param, a flat lowercased frozenset, and the request-behaviour tests now set it directly with no dual-write.

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -314,11 +314,28 @@ def mech_tool(self) -> str:
         """Get the selected mech tool."""
         return str(self.db.get_strict("mech_tool"))
 
+    @property
+    def selected_mechs(self) -> List[str]:
+        """Get the ChatUI-pinned mech addresses (lowercase). Empty means no pin."""
+        raw = self.db.get("selected_mechs", SERIALIZED_EMPTY_LIST)
+        if isinstance(raw, str):
+            raw = json.loads(raw)
```
bennyjo: The trader writes `selected_mechs` via consensus, so the input isn't untrusted — but a single bad write (malformed JSON, wrong shape) would raise from this property on every subsequent round until the db key is cleared. Worth wrapping with a try/except that logs and falls back to `[]`? A test case feeding `"not-json"` and asserting `[]` would lock it in.

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -314,11 +314,28 @@ def mech_tool(self) -> str:
...
+            a subset of mechs via `selected_mechs`, the result is further
+            restricted to that subset.
+        """
+        pinned = self.selected_mechs
         return [
-            info for info in self.mechs_info if self.mech_tool in info.relevant_tools
+            info
+            for info in self.mechs_info
+            if self.mech_tool in info.relevant_tools
+            and (not pinned or info.address.lower() in pinned)
```
bennyjo, line 348: When `selected_mechs` is non-empty but no pinned mech serves `mech_tool` (e.g. trader pins mech X but X doesn't advertise the chosen tool), this silently returns `[]` and the consumer retries with no diagnostic. The other empty-result paths write distinct `last_failure_reason` codes (`subgraph_unavailable`, `allowlist_not_configured`, `valid_mech_list_empty`, `no_overlap_with_valid_tools`) so the ChatUI can surface a specific remediation. A fifth code (e.g. `no_overlap_with_selected_mechs`) written at the consumer of `relevant_mechs_info` when empty + `selected_mechs` non-empty would close the loop. Happy to leave as a follow-up if you'd rather wire the diagnostic alongside the trader-side changes.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -297,8 +297,11 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:  # pragma: no cover
             "use_acn_for_delivers", kwargs, bool
         )
         self.irrelevant_tools: set = set(self._ensure("irrelevant_tools", kwargs, list))
-        self.ignored_mechs: FrozenSet[str] = frozenset(
-            self._ensure("ignored_mechs", kwargs, List[str])
+        self.valid_mechs: FrozenSet[str] = frozenset(
+            str(addr).lower() for addr in self._ensure("valid_mechs", kwargs, List[str])
+        )
+        self.valid_tools: FrozenSet[str] = frozenset(
```
bennyjo, line 303: Asymmetry: `valid_mechs` is lowercased on load (`str(addr).lower()`) but `valid_tools` isn't. Tool names in IPFS manifests are conventionally lowercase, but a typo like `Prediction-Online` in `service.yaml` would silently fail to intersect with `prediction-online` from the manifest and surface as `no_overlap_with_valid_tools` with no hint at the cause. Either lowercase here for symmetry, or add a one-line note in `UPGRADING.md` that `valid_tools` is matched case-sensitively. Happy to leave as-is if matching manifest casing is the intent.

bennyjo (commented):
Re-review after the flat-list refactor + `selected_mechs` feature. Three follow-up nits, all framed as questions.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -371,6 +374,15 @@ def validate_configuration(self) -> None:  # pragma: no cover
                         "priority_mech_address is required "
                         "when use_mech_marketplace is True and use_dynamic_mech_selection is False"
                     )
+                if (
+                    self.mech_marketplace_config.use_dynamic_mech_selection
+                    and not self.valid_mechs
+                ):
+                    self.context.logger.warning(
+                        "`valid_mechs` is empty while marketplace v2 dynamic "
```
LOCKhart07: Mirror the LOCKhart07 #2 fix on `valid_tools`. With v2 dynamic selection enabled and `valid_tools=[]`, `populate_tools` intersects each manifest with an empty set, so every mech ends up with empty `relevant_tools` and every round dies with `no_overlap_with_valid_tools` — same failure mode the `valid_mechs` warning here is guarding against, but with no boot-time signal. Add a parallel `if use_dynamic_mech_selection and not self.valid_tools:` warning here (or fold into one combined warning when either is empty).

packages/valory/skills/mech_interact_abci/graph_tooling/requests.py
```diff
@@ -155,6 +156,12 @@ def fetch_mechs_info(  # pragma: no cover
         self,
     ) -> MechsInfoFetcher:
         """Fetch mechs' information from the subgraph."""
+        if not self.params.valid_mechs:
+            self.context.logger.warning(
+                "`valid_mechs` is empty; skipping subgraph fetch."
```
LOCKhart07: This WARNING fires every round when `valid_mechs` is empty. The new boot-time warning in `validate_configuration` already surfaces the misconfig, and the `allowlist_not_configured` failure reason carries it forward to downstream consumers — so the per-round log is just noise. Suggest dropping it or downgrading to `debug`.

packages/valory/skills/mech_interact_abci/graph_tooling/requests.py
```diff
@@ -126,10 +126,11 @@ def fetch_mechs_info_batch(
         """Fetch a batch of mechs' information from the subgraph."""
         lookback_seconds = self.params.deliveries_lookback_days * SECONDS_PER_DAY
         block_timestamp_gt = int(time.time()) - lookback_seconds
+        # sorted: query string must be deterministic across agents in the same round
         query = mechs_info_query.substitute(
             first=QUERY_BATCH_SIZE,
             mechs_id_gt=mechs_id_gt,
-            ignored_mechs='", "'.join(self.params.ignored_mechs),
+            valid_mechs='", "'.join(sorted(self.params.valid_mechs)),
```
LOCKhart07, line 133: `sorted(...)` here silently fixes a pre-existing nondeterminism bug, not just a stylistic touch-up — `FrozenSet` iteration order depends on `PYTHONHASHSEED`, so the old `'", "'.join(self.params.ignored_mechs)` could produce different query strings on different agents in the same round, breaking FSM consensus on the http payload. Worth calling out in the commit message / PR description so this fix isn't lost as a side-effect — and so future readers know the old `ignored_mechs` code path was a latent FSM hazard.

packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -584,7 +584,18 @@ def get_priority_mech_address(self) -> Optional[str]:
             return priority_mech
 
         if self.params.use_mech_marketplace:
-            return self.mech_marketplace_config.priority_mech_address
+            static_priority = self.mech_marketplace_config.priority_mech_address
+            if (
+                static_priority
+                and self.params.valid_mechs
+                and static_priority.lower() not in self.params.valid_mechs
```
LOCKhart07, line 598: Code is fine, but it doesn't quite match the PR description. The description says "Static `priority_mech_address` rejected at runtime if not in `valid_mechs`", whereas the guard short-circuits on `self.params.valid_mechs` being truthy too — so when `valid_mechs=[]` (the legacy "not configured" state), the static priority is still honored. That's the sensible migration choice, but please tighten the PR description to something like "rejected when `valid_mechs` is configured (non-empty) and the priority is not in it".

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
```diff
@@ -507,6 +523,155 @@ def mock_fetch_mechs_info() -> Generator[None, None, List[MechInfo]]:
...
+            yield
+            return []
+
+        behaviour.fetch_mechs_info = mock_fetch_mechs_info  # type: ignore[method-assign]
+
+        _drive(behaviour.get_mechs_info())
+
+        assert behaviour._context.state.last_failure_reason == "valid_mech_list_empty"
+
+    def test_writes_no_overlap_when_metadata_disjoint_from_valid_tools(self) -> None:
```
LOCKhart07: The test name says "metadata disjoint from valid_tools" but the body uses `valid_tools=frozenset()` — that's the empty/unconfigured case, not disjoint. The branch under test fires equally for both, but a truly disjoint case (`valid_tools={"tool_x"}`, metadata=`["tool_a"]`) would actually prove the intersect-then-empty path works when both sides are non-empty. Either rename to `..._when_valid_tools_empty`, or add a second case with genuinely disjoint sets.

LOCKhart07 (dismissed):

packages/valory/skills/mech_interact_abci/states/base.py
DIvyaNautiyal07: Done in 53ba817. The property now wraps the db read in try/except for `(json.JSONDecodeError, TypeError, AttributeError)` and returns `[]` on malformed input. Two new tests in `TestSynchronizedData`: `test_selected_mechs_falls_back_on_malformed_json` (feeds `"not-json"`) and `test_selected_mechs_falls_back_on_wrong_shape` (feeds `json.dumps(42)`).

packages/valory/skills/mech_interact_abci/states/base.py
DIvyaNautiyal07: Wired in 53ba817 rather than left as follow-up. `MechRequestBehaviour.get_priority_mech_address` writes `no_overlap_with_selected_mechs` on `SharedState.last_failure_reason` when the dynamic-v2 path returns no candidate AND `synchronized_data.selected_mechs` is non-empty. New test in `TestGetPriorityMechAddress` covers it. Added to UPGRADING.md's failure-reason list.

packages/valory/skills/mech_interact_abci/models.py
DIvyaNautiyal07: Symmetric in 53ba817: `valid_tools` is lowercased on load (`str(tool).lower()`) and `populate_tools` lowercases manifest entries before the intersect. A `Prediction-Online` in `service.yaml` now matches a `prediction-online` manifest entry. UPGRADING.md notes the case-insensitive matching.

packages/valory/skills/mech_interact_abci/models.py
DIvyaNautiyal07: Done in 53ba817. Folded into the single boot-time warning so it lists whichever of `valid_mechs` / `valid_tools` is empty. No more silent dies-every-round path.

packages/valory/skills/mech_interact_abci/graph_tooling/requests.py
DIvyaNautiyal07: Dropped in 53ba817. The boot-time warning in `validate_configuration` and the `allowlist_not_configured` reason carry the diagnostic forward, so the per-round log is gone.

packages/valory/skills/mech_interact_abci/graph_tooling/requests.py
DIvyaNautiyal07: Good catch on the latent FSM hazard — calling it out in the PR description now. The old `ignored_mechs` path was vulnerable to `PYTHONHASHSEED`-dependent iteration order producing different POST bodies across agents in the same round; `sorted(...)` here gives deterministic strings.

packages/valory/skills/mech_interact_abci/behaviours/request.py
DIvyaNautiyal07: PR description tightened — "rejected when `valid_mechs` is configured (non-empty) and the priority is not in it." The empty-allowlist path (legacy / not-yet-migrated deployers) is intentionally a no-op so the bump doesn't regress them.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
DIvyaNautiyal07: Done in 53ba817. The empty-set test is renamed to `test_writes_no_overlap_when_valid_tools_empty`. A new `test_writes_no_overlap_when_metadata_disjoint_from_valid_tools` covers the genuinely-disjoint case (`valid_tools={"tool_x"}`, manifest=`["tool_a", "tool_b"]`) and asserts `mech.relevant_tools` is empty plus the reason fires.

bennyjo (approved):
LGTM — all review feedback from prior rounds landed cleanly.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -135,9 +135,12 @@ def populate_tools(
                 self.mech_tools_api.reset_retries()
                 continue
 
-            relevant_tools = set(res) - self.params.irrelevant_tools
+            metadata_tools = {str(t).lower() for t in res}
+            allowed_tools = (
+                metadata_tools - self.params.irrelevant_tools
+            ) & self.params.valid_tools
```
LOCKhart07, line 141: `irrelevant_tools` casing now diverges from `metadata_tools` here. Pre-PR, both sides of the subtract were case-sensitive identical strings (`{"Tool-A"} - {"Tool-A"} = {}` filtered correctly). Post-PR, `metadata_tools` is lowercased but `irrelevant_tools` is loaded as-is in `models.py:299` — so `{"tool-a"} - {"Tool-A"} = {"tool-a"}` silently fails to filter.

The bundled `skill.yaml` defaults are all lowercase so this isn't broken in practice, but any deployer who overlays `irrelevant_tools` with mixed casing will silently regress.

Suggest lowercasing on load to mirror `valid_tools` / `valid_mechs` / manifest:

```python
self.irrelevant_tools: set = {
    str(t).lower() for t in self._ensure("irrelevant_tools", kwargs, list)
}
```

Non-blocking — approving since the defaults make this latent.

LOCKhart07 (approved):

DhairyaPatel7 (approved):

mariapiamo (commented):
1. `selected_mechs` does not narrow the exposed tool list before trader selects a tool.

`selected_mechs` is applied only in `SynchronizedData.relevant_mechs_info`, after `mech_tool` has already been selected. However, `SynchronizedData.mech_tools` still returns the union of tools across all discovered mechs, ignoring `selected_mechs`.

This can produce an invalid flow:

```text
valid_mechs = [A, B]
valid_tools = [tool_x, tool_y]

A supports tool_x
B supports tool_y

selected_mechs = [B]

mech_tools still exposes [tool_x, tool_y]
trader selects tool_x
relevant_mechs_info filters to selected_mechs=[B]
no mech remains for tool_x
At that point get_priority_mech_address() can return None, and request preparation can keep retrying because wait_for_condition_with_sleep() has no timeout by default.

Suggested fix: when selected_mechs is set, mech_tools should expose only tools supported by those selected mechs, or the flow should fail before request construction with an explicit no-candidate reason.

2. last_failure_reason is not synchronized.
The PR documents SharedState.last_failure_reason as something trader/ChatUI can consume, but it is only local SharedState. It is not part of the round payload, and FailedMechInformationRound does not write it to synchronized data.

Suggested fix: include the failure reason in the MechInformationRound payload and persist it in synchronized data if another skill is expected to consume it.

3. Dynamic V2 still falls back to a penalized mech.
Current dynamic selection still does:

first ranked mech not penalized,
else fallback to priority_mech_address
So if all relevant valid mechs are temporarily penalized, the code can still select the top penalized mech.

Suggested fix: if all relevant valid mechs are penalized, fail fast instead of falling back to a penalized mech.

4. Static priority can bypass the allowlist when valid_mechs is empty.
The static priority guard only rejects the configured priority mech when valid_mechs is non-empty. With the default valid_mechs = [], a configured marketplace priority_mech_address can still be used.

If valid_mechs is meant to be the trust boundary, empty should mean “not configured / no eligible marketplace mechs”, not “allow static priority”.

Suggested fix: document that static priority mode intentionally bypasses the allowlist.

PR #79, opened 2026-05-12
fix(mech_interact_abci): honor selected_mechs in mech_tools + never fall back to penalized mechs

DIvyaNautiyal07:
## Summary

Follow-up to PR #78 addressing [mariapiamo's post-merge review](https://github.com/valory-xyz/mech-interact/pull/78#pullrequestreview-4272803892). Three small fixes that close real gaps in the allowlist behavior.

1. **`mech_tools` now narrows by `selected_mechs`.** Previously the property returned the union of `relevant_tools` across *all* discovered mechs, ignoring the pin. A consumer could pick a tool that no pinned mech serves and the round would dead-end at request prep. Mirrors the existing `relevant_mechs_info` filter so the tool-selection API is internally consistent.

2. **Dynamic-V2 no longer falls back to a penalized mech.** The old code fell back to `priority_mech_address` (the top of the ranked list) when every non-penalized candidate was missing — but the top of the ranked list IS penalized in that case. Per spec: "if all filtered valid candidates are under active temporary local penalty, fail fast with `no_non_penalized_valid_mech`." Now: `get_priority_mech_address` returns `None` and writes the specific reason. Precedence: `no_non_penalized_valid_mech` wins when `ranked` is non-empty (pinned mech exists but penalized); `no_overlap_with_selected_mechs` wins when `ranked` is empty and a pin is set.

3. **Static-priority guard documented.** Added an in-code comment explaining the intentional short-circuit when `valid_mechs` is empty (legacy deployers not yet migrated; the trust boundary kicks in only when `valid_mechs` is populated).

## Deferred

Mariapiamo's review item #2 (synchronize `last_failure_reason` through `SynchronizedData` rather than per-agent `SharedState`) is intentionally not addressed here. For single-agent Pearl deployments, trader's ChatUI handler reads the same `SharedState` so consumption works. Multi-agent consistency tracked as a separate follow-up.

## Test plan

- [x] 384 unit tests pass (3 new: `mech_tools` narrow / no-op cases, all-pinned-and-penalized precedence)
- [x] Static checks clean on touched files: mypy, flake8, pylint, darglint, black, isort
- [x] `autonomy packages lock --check` passes; `autonomy push-all` published the new skill hash to IPFS

## Breaking change

Documented in `UPGRADING.md` under `Unreleased`. The dynamic-V2 penalize-fallback behavior is gone — services relying on "if all penalized, try the top one anyway" will now fail-fast instead. Adjust penalty windows accordingly.

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/request.py
@@ -564,6 +564,7 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
 
     def get_priority_mech_address(self) -> Optional[str]:
         """Get the priority mech's address. Warning: the result is based on the time of access."""
+        self.shared_state.last_failure_reason = None
         if (
             self.should_use_marketplace_v2()
             and self.mech_marketplace_config.use_dynamic_mech_selection
@@ -979,6 +980,25 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
         """Do the action."""
 
         if not self._mech_requests:
+            with self.context.benchmark_tool.measure(self.behaviour_id).local():
+                payload = MechRequestPayload(
+                    self.context.agent_address,
+                    self.matching_round.auto_round_id(),
+                    None,
+                    None,
+                    self.params.mech_chain_id,
+                    self.synchronized_data.safe_contract_address,
+                    SERIALIZED_EMPTY_LIST,
+                    SERIALIZED_EMPTY_LIST,
+                )
+            yield from self.finish_behaviour(payload)
+            return
+
+        if not self.priority_mech_address:
+            self.context.logger.warning(
+                "No priority mech available this round; skipping request "
+                f"(last_failure_reason={self.shared_state.last_failure_reason!r})."
+            )
             with self.context.benchmark_tool.measure(self.behaviour_id).local():
                 payload = MechRequestPayload(
                     self.context.agent_address,
--- a/packages/valory/skills/mech_interact_abci/behaviours/request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/request.py
@@ -585,6 +585,8 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
                 self.shared_state.last_failure_reason = "no_non_penalized_valid_mech"
             elif self.synchronized_data.selected_mechs:
                 self.shared_state.last_failure_reason = "no_overlap_with_selected_mechs"
+            else:
+                self.shared_state.last_failure_reason = "no_overlap_with_selected_tool"
             return None
 
         if self.params.use_mech_marketplace:
--- a/packages/valory/skills/mech_interact_abci/behaviours/request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/request.py
@@ -605,6 +605,9 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
                     f"Configured priority_mech_address {static_priority} is not in "
                     f"`valid_mechs`; skipping mech request."
                 )
+                self.shared_state.last_failure_reason = (
+                    "static_priority_not_in_valid_mechs"
+                )
                 return None
             return static_priority
 
@@ -976,21 +979,25 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
 
         return False
 
+    def _build_skip_payload(self) -> MechRequestPayload:
+        """Build the no-op payload used by both skip paths in `async_act`."""
+        return MechRequestPayload(
+            self.context.agent_address,
+            self.matching_round.auto_round_id(),
+            None,
+            None,
+            self.params.mech_chain_id,
+            self.synchronized_data.safe_contract_address,
+            SERIALIZED_EMPTY_LIST,
+            SERIALIZED_EMPTY_LIST,
+        )
+
     def async_act(self) -> Generator:  # pragma: no cover
         """Do the action."""
 
         if not self._mech_requests:
             with self.context.benchmark_tool.measure(self.behaviour_id).local():
-                payload = MechRequestPayload(
-                    self.context.agent_address,
-                    self.matching_round.auto_round_id(),
-                    None,
-                    None,
-                    self.params.mech_chain_id,
-                    self.synchronized_data.safe_contract_address,
-                    SERIALIZED_EMPTY_LIST,
-                    SERIALIZED_EMPTY_LIST,
-                )
+                payload = self._build_skip_payload()
             yield from self.finish_behaviour(payload)
             return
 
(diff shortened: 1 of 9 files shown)


mariapiamo (commented):
1. High: The new “fail fast” path still hangs in request preparation. get_priority_mech_address() now returns None when all ranked mechs are penalized or selected mechs have no overlap: request.py (line 571). But async_act() still calls _prepare_safe_tx() unconditionally when there are mech requests: request.py (line 994). _prepare_safe_tx() then waits forever on _fetch_and_validate_payment_type: request.py (line 941), because _mech_mm_contract_interact() returns False forever when priority_mech_address is empty: request.py (line 645), and wait_for_condition_with_sleep() has no timeout by default: base.py (line 281). The PR changes the selected address, but not the control flow needed to exit the round. Add an early terminal/skip payload when no priority mech is available before _prepare_safe_tx() starts.

2. Medium: last_failure_reason can remain stale after a later successful selection. The new failure paths write last_failure_reason, but when a later call finds a valid priority_mech, it returns immediately without clearing the old reason: request.py (line 573). If the local trader/ChatUI reads this field, it can display no_non_penalized_valid_mech or no_overlap_with_selected_mechs after the next request is already valid. Clear the request-level failure reason in setup() or before returning a successful priority mech.

3. Low: selected_mechs malformed-shape handling is incomplete. The docstring says wrong shapes return an empty list, but JSON strings and JSON objects do not raise here: base.py (line 327). For example, "0xabc" becomes a list of characters and {"0xabc": true} becomes ["0xabc"]. Validate raw is a list after json.loads; otherwise return [].

packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -568,30 +568,31 @@ def get_priority_mech_address(self) -> Optional[str]:
...
+            # Distinguish (a) every relevant candidate is penalized vs
+            # (b) the user's pin yielded no overlap with the chosen tool.
+            # Per spec, each is a fail-fast terminal — never return a
+            # penalized mech as a fallback.
+            self.context.logger.warning("No whitelisted priority mech found!")
+            if ranked:
+                self.shared_state.last_failure_reason = "no_non_penalized_valid_mech"
+            elif self.synchronized_data.selected_mechs:
+                self.shared_state.last_failure_reason = "no_overlap_with_selected_mechs"
+            return None
```
LOCKhart07, line 591: Non-blocking: there's a silent third path here. When `ranked` is empty AND `selected_mechs` is empty (mechs are populated but none serves the chosen `mech_tool`), we `return None` without setting `last_failure_reason`. `MechInformationBehaviour` only writes `no_overlap_with_valid_tools` when *no* mech serves *any* valid_tool — not when no mech serves the specifically selected tool — so trader's ChatUI handler would see a `None` reason for what is actually a recognisable failure. Pre-existing gap from PR #78's design, but worth a follow-up given this PR explicitly closes diagnostic gaps. Either add a `no_overlap_with_selected_tool` (or similar) reason here, or tighten `MechInformationBehaviour` to check `mech_tool` specifically.

packages/valory/skills/mech_interact_abci/tests/test_request_behaviour.py
```diff
@@ -284,7 +284,7 @@ def test_marketplace_v2_dynamic_all_penalized_returns_fallback(
 
         mock_synced = MagicMock()
         mock_synced.ranked_mechs_addresses = ["0xpenalized1", "0xpenalized2"]
-        mock_synced.priority_mech_address = "0xfallback"
+        mock_synced.selected_mechs = []
```
LOCKhart07, line 287: Non-blocking: `mock_synced.priority_mech_address = "0xpenalized1"` is dead — the new `get_priority_mech_address` no longer reads `synchronized_data.priority_mech_address` (the fallback path that consumed it was removed). Safe to drop for clarity. Same applies to line 324 in the next test, though there the value coincides with `ranked_mechs_addresses[0]` so it's at least self-consistent.

packages/valory/skills/mech_interact_abci/behaviours/request.py
DIvyaNautiyal07: Fixed in 80f4775. Added an `else` branch that writes `no_overlap_with_selected_tool` when neither `ranked` is truthy nor `selected_mechs` is set — closes the silent-None path you described. New reason is also documented in UPGRADING.md and there's a new test `test_v2_dynamic_writes_no_overlap_with_selected_tool_when_no_pin` locking it in.

packages/valory/skills/mech_interact_abci/tests/test_request_behaviour.py
DIvyaNautiyal07: Dropped in 80f4775. Same treatment as the line 287 one — `get_priority_mech_address` no longer reads `synchronized_data.priority_mech_address` on either path.

LOCKhart07 (dismissed):
All prior review items resolved cleanly — verified the four fixes against the final state of `request.py`, `mech_info.py`, `states/base.py`, and the new tests. The `last_failure_reason` reset, `async_act` early-exit, `selected_mechs` shape guard, `pinned_mechs_offline` check, and `mech_tools` narrowing all land as described and the new tests are falsifiable.

Three non-blocking observations, all follow-uppable:

1. **Static-V2 valid_mechs guard stays silent.** `get_priority_mech_address` returns `None` in `behaviours/request.py:592-606` when a configured `static_priority` is not in `valid_mechs`, but writes no `last_failure_reason`. The new `async_act` skip path will log `last_failure_reason=None`, so trader-side ChatUI can't distinguish "operator misconfigured `priority_mech_address`" from any other failure. The rest of this PR is explicitly about closing those diagnostic gaps; a reason like `static_priority_not_in_valid_mechs` would round out the set. Pre-existing from #78, not a regression here.

2. **`SynchronizedData.selected_mechs` docstring drift.** `states/base.py:319-323` claims malformed values return an empty list "with a warning," but no warning is emitted (and `SynchronizedData` has no logger). Touching the property anyway — easy one-line docstring fix.

3. **Skip-payload duplication in `async_act`.** The new empty-`priority_mech_address` branch emits a `MechRequestPayload` that is byte-for-byte identical to the existing `not self._mech_requests` skip payload. A small `_emit_skip_payload()` helper would remove the duplication. Style only.

Nothing blocking — happy to land as-is and follow up on #1 separately.

mariapiamo (commented):
Selected mech visible but no valid tools gives no clear failure reason

They added pinned_mechs_offline, but only for this case:

selected_mechs has no overlap with fetched mech_info
There is another possible no-candidate case:

valid_mechs = [A, B]
valid_tools = [tool_x]

A supports tool_x
B is selected_mechs
B is visible in subgraph
B has no valid tools
Now:

mech_info succeeds
pinned_mechs_offline does not fire because B is visible
mech_tools narrows to selected_mechs = [B]
mech_tools becomes empty
trader cannot select a tool
no failure reason is written
So the PR should also fail fast when selected mechs are visible but none of them expose any allowed tool.

LOCKhart07 (approved):

DhairyaPatel7 (approved):

mariapiamo (approved):

PR #80, opened 2026-05-13
chore(mech_interact_abci): drop irrelevant_tools, valid_tools is the single allowlist

DIvyaNautiyal07:
## Summary

Following the PR #79 review thread with @bennyjo: keeping both `irrelevant_tools` (denylist) and `valid_tools` (allowlist) is redundant. Operators can express any denylist intent by simply omitting tools from `valid_tools`. This PR collapses the per-mech filter from `(metadata_tools - irrelevant_tools) & valid_tools` to `metadata_tools & valid_tools`.

## Changes
- `models.py`: drop `self.irrelevant_tools` and its `_ensure("irrelevant_tools", ...)` call from `MechParams.__init__`.
- `behaviours/mech_info.py`: drop the subtraction step in `populate_tools` (see [the previous line at `mech_info.py:140-141`](https://github.com/valory-xyz/mech-interact/blob/v0.31.0/packages/valory/skills/mech_interact_abci/behaviours/mech_info.py#L140-L141) for the old form).
- `skill.yaml`: drop the default `irrelevant_tools` list.
- Tests: drop the dedicated `test_irrelevant_tools_filtered_before_intersect`; simplify `test_populates_tools_from_http_response` to exercise the `valid_tools`-only path.
- `UPGRADING.md`: document the breaking change and the migration path.

## Where the prod denylist lived
The production `IRRELEVANT_TOOLS` env var template lives in operate-app at [`frontend/constants/serviceTemplates/service/trader.ts:124-128`](https://github.com/valory-xyz/olas-operate-app/blob/main/frontend/constants/serviceTemplates/service/trader.ts#L124-L128) (Pearl/Gnosis) and [`:271-276`](https://github.com/valory-xyz/olas-operate-app/blob/main/frontend/constants/serviceTemplates/service/trader.ts#L271-L276) (Polystrat/Polygon). Pearl + operate-app need to drop those `IRRELEVANT_TOOLS` templates in a follow-up; trader's seeded `valid_tools` will be curated down to the effective set (`prediction-offline`, `superforcaster`, `factual_research` on Gnosis; `superforcaster` on Polygon) in the companion trader PR.

## Test plan
- [x] Unit tests pass (`uv run pytest packages/valory/skills/mech_interact_abci/`).
- [x] `tomte tox -e black-check -e isort-check -e flake8 -e mypy -e darglint` clean.
- [x] `autonomy packages lock --check` succeeds.
- [x] `autonomy push-all` published the new hash to IPFS.
- [ ] Companion trader PR to land after this merges + `v0.32.0` tag is cut.

New mech_interact_abci hash: \`bafybeihds6vqk3gsevtsdsowz56jdnbbhc6gnswwd25opm6rrj32dhsmou\`.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,6 +5,11 @@ Below, we describe the additional manual steps required to upgrade between diffe
 ## Unreleased (built with `open-aea@2.2.3` and `open-autonomy@0.21.20`)
 
 #### Breaking Changes
+- The `irrelevant_tools` blocklist has been removed. Any overlay that still
+  sets `irrelevant_tools` will be silently ignored. Curate `valid_tools`
+  as the single tool allowlist instead: anything previously in
+  `irrelevant_tools` must be omitted from `valid_tools`. Per-mech
+  `relevant_tools` is now just `metadata_tools & valid_tools`.
 - The `ignored_mechs` blocklist has been removed. Any overlay
   (`aea-config.yaml`, `service.yaml`) that still sets `ignored_mechs` will be
   silently ignored. Replace it with the new `valid_mechs` allowlist.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -101,6 +101,8 @@ Below, we describe the additional manual steps required to upgrade between diffe
     - `mechs_info`: All the mechs' information.
     - `relevant_mechs_info`: The mechs' information that are relevant to the user, 
       i.e., include tools which are not in the `irrelevant_tools` set.
+      (Note: `irrelevant_tools` was removed in the Unreleased section above;
+      from that version on, the filter is `metadata_tools & valid_tools`.)
     - `mech_tools`: The set of all the mechs' tools.
     - `priority_mech`: The dynamically picked priority mech.
     - `priority_mech_address`: The address of the dynamically picked priority mech.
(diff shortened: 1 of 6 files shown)


LOCKhart07 (dismissed):

mariapiamo (commented):
1. Medium: stale irrelevant_tools configs are silently ignored.
PR 80 removes irrelevant_tools from Params and uses only:
metadata_tools & valid_tools
See packages/valory/skills/mech_interact_abci/behaviours/mech_info.py:138-141.
This is coherent with the simplified design, but the migration is fail-open if an operator still has an old overlay with irrelevant_tools and a broad valid_tools. The removed denylist will no longer protect anything. I would prefer at least a startup warning, or even a hard failure, when irrelevant_tools is still present in kwargs.

2. Low: UPGRADING.md still has old wording that can confuse readers.
The new section says irrelevant_tools is removed, but older text still says relevant_mechs_info means tools “not in the irrelevant_tools set” and that "irrelevant_tools are now defined in this skill": UPGRADING.md:102-120. Historical sections can remain, but this now reads contradictory.

DhairyaPatel7 (approved):

LOCKhart07 (approved):

UPGRADING.md
```diff
@@ -5,6 +5,11 @@ Below, we describe the additional manual steps required to upgrade between diffe
 ## Unreleased (built with `open-aea@2.2.3` and `open-autonomy@0.21.20`)
 
 #### Breaking Changes
+- The `irrelevant_tools` blocklist has been removed. Any overlay that still
+  sets `irrelevant_tools` will be silently ignored. Curate `valid_tools`
```
bennyjo, line 9: Soft question — did you sanity-check this "silently ignored" claim by booting an agent with a leftover `irrelevant_tools:` in an overlay before publishing the new IPFS hash? It's plausible (`_ensure` only enforces *required* keys, unconsumed kwargs flow through to `super().__init__`, and the same precedent worked for the `ignored_mechs` removal), so happy to leave as-is if you've already confirmed. Just flagging because if open-autonomy ever tightens base-param validation, this is the line that becomes wrong.

bennyjo (approved):
Approving — diff matches the description exactly and removes the redundancy cleanly. One soft question on the migration note, inline.

PR #89, opened 2026-05-19
chore(mech_interact_abci): drop valid_tools allowlist; suitability is the consumer's job

jmoreira-valory:
## Summary

- Per-mech `relevant_tools` now carries the full IPFS manifest tool set; the skill no longer pre-filters against an operator-curated `valid_tools` allowlist.
- The `valid_tools` param is removed entirely from `mech_interact_abci` -- consumers (e.g. trader's `decision_maker_abci`) become responsible for tool-suitability decisions.
- Two `last_failure_reason` strings renamed to reflect the new trigger conditions: `no_overlap_with_valid_tools` -> `no_tools_in_manifests`, `pinned_mechs_no_valid_tools` -> `pinned_mechs_no_tools`.

## Why

`relevant_tools` mixing transport (what the mech declares) with policy (what the operator authorises) made it impossible for downstream consumers to reason about which prediction tools the marketplace actually offers vs. which the operator allowlisted. Splitting the concern cleanly lets the consumer apply richer, content-based filters (e.g. trader's new `is_prediction_tool` classifier reads the actual manifest `toolMetadata` to decide).

## Cross-fleet impact

Audit confirmed only **trader** consumes `mech.relevant_tools` (see `decision_maker_abci/behaviours/tool_selection.py`). market-resolver depends on `mech_interact_abci` for transport but hardcodes a single tool; market-creator does not compose the skill. The trader companion PR replaces the allowlist with a tool-suitability classifier.

## Test plan

- [x] `tomte tox -e isort,black,flake8,mypy,darglint` all green
- [x] Pylint clean on touched files (10/10)
- [x] `autonomy analyse docstrings --update` no diff
- [x] `autonomy packages lock` produces a single hash bump (`mech_interact_abci`)
- [x] `pytest packages/valory/skills/mech_interact_abci/tests/` 450 passed at 100% coverage
- [ ] `mech-interact` CI green
- [ ] Downstream trader PR builds and passes against the new CID

## Breaking changes

Documented in `UPGRADING.md`:

1. The `valid_tools` param on `mech_interact_abci` is removed; any overlay that still sets it will fail to load. Move the override into the consumer skill that needs it.
2. `last_failure_reason` values renamed: `no_overlap_with_valid_tools` -> `no_tools_in_manifests`, `pinned_mechs_no_valid_tools` -> `pinned_mechs_no_tools`. Consumers that switch on these strings must update.

## Test churn

- `TestAllowlistIntersect` class deleted (premise gone).
- `test_writes_no_overlap_when_metadata_disjoint_from_valid_tools` deleted (unreachable now).
- `test_writes_no_overlap_when_valid_tools_empty` repurposed to `test_writes_no_tools_in_manifests_when_all_manifests_empty`.
- `test_writes_pinned_mechs_no_valid_tools_when_pinned_visible_but_no_tools` renamed and adapted.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,6 +5,22 @@ Below, we describe the additional manual steps required to upgrade between diffe
 ## Unreleased (built with `open-aea@2.2.6` and `open-autonomy@0.21.22`)
 
 #### Breaking Changes
+- The `valid_tools` operator allowlist has been removed from
+  `mech_interact_abci`. The skill no longer filters per-mech
+  `relevant_tools` against an operator-curated set; each mech's
+  `relevant_tools` now carries its full IPFS manifest tool list as-is.
+  Tool-suitability decisions are the consumer's responsibility (e.g.
+  `decision_maker_abci` in trader runs an `is_prediction_tool`
+  classifier and may still keep its own `valid_tools` safety net).
+  Any overlay that still sets `valid_tools` on `mech_interact_abci`
+  will fail to load because the param is no longer declared. Move the
+  override into the consumer skill that needs it (e.g.
+  `decision_maker_abci`).
+- The `last_failure_reason` value `no_overlap_with_valid_tools` is
+  renamed to `no_tools_in_manifests` (the only way the trigger
+  condition fires now is when every fetched manifest is empty), and
+  `pinned_mechs_no_valid_tools` is renamed to `pinned_mechs_no_tools`.
+  Consumers that switch on these strings must update accordingly.
 - The `irrelevant_tools` blocklist has been removed. Any overlay that still
   sets `irrelevant_tools` will be silently ignored. Curate `valid_tools`
   as the single tool allowlist instead: anything previously in
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -17,15 +17,19 @@ Below, we describe the additional manual steps required to upgrade between diffe
   override into the consumer skill that needs it (e.g.
   `decision_maker_abci`).
 - The `last_failure_reason` value `no_overlap_with_valid_tools` is
-  renamed to `no_tools_in_manifests` (the only way the trigger
-  condition fires now is when every fetched manifest is empty), and
-  `pinned_mechs_no_valid_tools` is renamed to `pinned_mechs_no_tools`.
-  Consumers that switch on these strings must update accordingly.
-- The `irrelevant_tools` blocklist has been removed. Any overlay that still
-  sets `irrelevant_tools` will be silently ignored. Curate `valid_tools`
-  as the single tool allowlist instead: anything previously in
-  `irrelevant_tools` must be omitted from `valid_tools`. Per-mech
-  `relevant_tools` is now just `metadata_tools & valid_tools`.
+  renamed to `no_usable_tools_in_mechs`, and
+  `pinned_mechs_no_valid_tools` is renamed to
+  `pinned_mechs_no_usable_tools`. Both reasons now cover the broader
+  trigger: a mech ends up with empty `relevant_tools` when its
+  manifest is empty, permanently failing, or retries-exhausted. The
+  rename reflects this — consumers that switch on the old strings
+  must update accordingly.
+- The `irrelevant_tools` blocklist has been removed. Any overlay that
+  still sets `irrelevant_tools` will be silently ignored. With this
+  release's removal of `valid_tools` too, mech_interact_abci no longer
+  filters tools at the transport layer at all — each mech's
+  `relevant_tools` carries its full IPFS manifest tool list, and any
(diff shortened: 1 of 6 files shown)


UPGRADING.md
```diff
@@ -5,6 +5,22 @@ Below, we describe the additional manual steps required to upgrade between diffe
...
+  `mech_interact_abci`. The skill no longer filters per-mech
+  `relevant_tools` against an operator-curated set; each mech's
+  `relevant_tools` now carries its full IPFS manifest tool list as-is.
+  Tool-suitability decisions are the consumer's responsibility (e.g.
+  `decision_maker_abci` in trader runs an `is_prediction_tool`
+  classifier and may still keep its own `valid_tools` safety net).
+  Any overlay that still sets `valid_tools` on `mech_interact_abci`
+  will fail to load because the param is no longer declared. Move the
+  override into the consumer skill that needs it (e.g.
+  `decision_maker_abci`).
```
DIvyaNautiyal07, line 18: Nice-to-have: the rest of the Unreleased section still references `valid_tools` from prior bullets, which now reads a bit inconsistent alongside this removal:

- L25-28 (the `irrelevant_tools` migration advice) tells operators to use `valid_tools` and says `relevant_tools = metadata_tools & valid_tools` - the param doesn't exist after this PR.
- L55-59 (`New features`) still announces `valid_tools` as a new param with intersect semantics.
- L67 and L70-71 still list `no_overlap_with_valid_tools` and `pinned_mechs_no_valid_tools` in the failure-reason inventory (this PR renames them).
- L121 (the historical v0.21.3->v0.22.0 footnote) still says the filter is `metadata_tools & valid_tools`.

Since the Unreleased block becomes one release's notes at cut time, a quick sweep to either remove the now-superseded bullets or update the strings would make the release notes internally consistent. Not blocking.

packages/valory/skills/mech_interact_abci/tests/test_mech_info_behaviour.py
```diff
@@ -91,14 +91,12 @@ def mock_get_http_response(**kwargs: Any) -> Any:
 def _setup_api(
     behaviour: MechInformationBehaviour,
     valid_mechs: Optional[FrozenSet[str]] = None,
```
DIvyaNautiyal07: Nice-to-have: with `valid_tools` removed from this helper, the remaining `valid_mechs` kwarg is unused at every call site in this file (all 17 calls pass it as the default). The two tests that actually need a populated `valid_mechs` (around lines 539 and 557) set `behaviour._context.params.valid_mechs` directly and bypass the helper. Since you're already trimming the signature, could drop `valid_mechs` too and hardcode `frozenset()` inside. Optional.

DIvyaNautiyal07 (dismissed):

UPGRADING.md
```diff
@@ -5,6 +5,22 @@ Below, we describe the additional manual steps required to upgrade between diffe
...
+- The `last_failure_reason` value `no_overlap_with_valid_tools` is
+  renamed to `no_tools_in_manifests` (the only way the trigger
+  condition fires now is when every fetched manifest is empty), and
```
LOCKhart07: Nit: the parenthetical narrows the trigger more than the code does. `no_tools_in_manifests` fires from `not any(mech.relevant_tools for mech in mech_info)` (mech_info.py:177), which is true whenever every mech ended up in `_failed_mechs` — that includes the permanent-error quarantine path and the retries-exhausted path, not only empty manifests. A consumer that switches on this string and infers "the IPFS manifests are empty" will be wrong some of the time when the gateway just returned 500s. Either soften the wording to "no mech has usable tools (manifests empty or unreachable)" or rename the reason to reflect the broader condition. Non-blocking.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -196,10 +195,10 @@ def get_mechs_info(
                 for mech in mech_info
             ):
                 self.context.logger.warning(
```
LOCKhart07, line 198: Same misrepresentation as the renamed `pinned_mechs_no_tools`: a visible pinned mech can land here with `relevant_tools` empty for any of three reasons — empty manifest, permanent error, retries exhausted — not just an empty manifest. The new wording implies we have the manifest content. Suggest something like `"...are visible but have no usable tools (manifest empty or unreachable)."` Non-blocking.

LOCKhart07 (dismissed):

DIvyaNautiyal07 (approved):
Re-review LGTM. All four prior threads resolved correctly: UPGRADING.md sweep is internally consistent, `_setup_api` is trimmed, and the failure-reason rename + softened log wording covers the broader trigger LOCKhart07 flagged.

LOCKhart07 (approved):

PR #93, opened 2026-06-17
feat(offchain): offchain dispatch branch — config foundation [WS-3]

DhairyaPatel7:
## WS-3 — mech-interact offchain dispatch (the plan's "dominant cost")

Part of the Marketplace/API offchain migration (Thread 1, requester side). Targets mech#454's server contract (structured 402, `/send_signed_requests`, `/fetch_offchain_info`, bare-file CID) and pairs with mech-client#241.

**This PR lands the foundation; the behaviour + FSM work follows (scoped below).** Ships dark — `use_offchain` defaults False, so the on-chain MechMarketplace path is bit-for-bit unchanged.

### In this PR (foundation)
- **`MechMarketplaceConfig`**: `use_offchain` (default False) + `offchain_url`, with validation (offchain needs a static URL *or* dynamic selection to discover the mech's URL).
- **`states/base`**: the offchain `last_failure_reason` constants — `offchain_all_failed`, `offchain_402_insufficient`, `offchain_503_all_mechs`, `offchain_timeout_all_mechs`.
- **`skill.yaml`**: the two config args.
- **Tests** for the config (default-off / static-url / dynamic / no-url-no-dynamic error). 38 model tests green; black-clean; relocked `mech_interact_abci` only (no third-party churn).

### Full design (for the follow-ups)
When `use_offchain` is true, the request/response behaviours diverge from the on-chain path at these points (mapped):

**Request side** (`behaviours/request.py`):
1. **Selection** (`get_priority_mech_address`, ~565): resolve the mech's HTTP URL — from `offchain_url`, or discovered per-mech from the on-chain manifest (`Service.metadata_str`) under dynamic selection.
2. **Build/dispatch** (`_build_request_data`/`_prepare_safe_tx`, ~901/947): instead of uploading to IPFS + building the multisend tx, compute the request CID locally (bare-file, matching the mech), sign the `request_id`, and **HTTP POST** to the mech's `/send_signed_requests` (`get_http_response`).
3. **402 / failover**: parse the structured 402 → set `offchain_402_insufficient`; on timeout/503 → re-sign for the next ranked mech **at the same on-chain nonce** (client-side retry, *not* step-in) → `offchain_503_all_mechs` / `offchain_timeout_all_mechs` / `offchain_all_failed` when exhausted.
4. **Payload** (`async_act`, ~995): record the pending responses for polling; emit no on-chain tx.

**Response side** (`behaviours/response.py`, `_get_response` ~468): point the response API at the mech's `/fetch_offchain_info` and poll, instead of the IPFS gateway.

**Consensus model (to confirm):** the offchain POST is a per-agent side-effect; the mech dedupes by `request_id` (same as #454's replay handling). Worth a team decision vs. a single-keeper POST.

### Follow-ups (checklist)
- [ ] Request behaviour: offchain sign + POST + 402/failover (gated on `use_offchain`).
- [ ] Response behaviour: offchain poll branch.
- [ ] Per-mech HTTP URL discovery from the on-chain manifest.
- [ ] **Composition FSM routing** (in the consuming agents — trader/market-creator, WS-4-adjacent): route the offchain request to the response round **skipping `TransactionSettlement`**. This is cross-repo and the main integration step.
- [ ] CID-path parity with the mech's bare-file `compute_cidv1` (benny's note on mech#454).

### DIFF
--- a/packages/valory/skills/mech_interact_abci/models.py
+++ b/packages/valory/skills/mech_interact_abci/models.py
@@ -232,11 +232,28 @@ class MechMarketplaceConfig:
     priority_mech_service_id: int = 975
     requester_staking_instance_address: Optional[str] = NULL_ADDRESS
     use_dynamic_mech_selection: bool = True
+    # Off-chain dispatch (ships dark). False = today's on-chain MechMarketplace
+    # request, bit-for-bit unchanged. True engages the off-chain branch: the
+    # request is HTTP-POSTed to the mech and the response is polled, instead of
+    # being submitted/settled on-chain. ``offchain_url`` is the static fallback
+    # endpoint; when unset the URL is discovered per-mech from the on-chain
+    # manifest (see OFFCHAIN follow-up).
+    use_offchain: bool = False
+    offchain_url: Optional[str] = None
 
     def __post_init__(self) -> None:
         """Validate configuration after initialization."""
         if self.response_timeout <= 0:
             raise ValueError("response_timeout must be positive")
+        if (
+            self.use_offchain
+            and not self.offchain_url
+            and not self.use_dynamic_mech_selection
+        ):
+            raise ValueError(
+                "use_offchain requires either offchain_url or "
+                "use_dynamic_mech_selection (to discover the mech's URL)"
+            )
 
 
 class MechParams(BaseParams):
--- a/packages/valory/skills/mech_interact_abci/states/base.py
+++ b/packages/valory/skills/mech_interact_abci/states/base.py
@@ -42,14 +42,6 @@ from packages.valory.skills.transaction_settlement_abci.rounds import (
 
 SERIALIZED_EMPTY_LIST = "[]"
 METADATA_FIELD = "metadata"
-
-# Off-chain dispatch `last_failure_reason` values, surfaced so operators and
-# downstream skills can branch on a stable label when the off-chain path
-# exhausts its options (see the off-chain request/response behaviours).
-OFFCHAIN_ALL_FAILED = "offchain_all_failed"
-OFFCHAIN_402_INSUFFICIENT = "offchain_402_insufficient"
-OFFCHAIN_503_ALL_MECHS = "offchain_503_all_mechs"
-OFFCHAIN_TIMEOUT_ALL_MECHS = "offchain_timeout_all_mechs"
 BLOCK_TIMESTAMP_FIELD = "blockTimestamp"
 METADATA_PREFIX_SIZE = 2
 MACHINE_EPS = 1e-9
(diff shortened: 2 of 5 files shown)


packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -232,11 +232,28 @@ class MechMarketplaceConfig:
     priority_mech_service_id: int = 975
     requester_staking_instance_address: Optional[str] = NULL_ADDRESS
     use_dynamic_mech_selection: bool = True
+    # Off-chain dispatch (ships dark). False = today's on-chain MechMarketplace
+    # request, bit-for-bit unchanged. True engages the off-chain branch: the
+    # request is HTTP-POSTed to the mech and the response is polled, instead of
+    # being submitted/settled on-chain. ``offchain_url`` is the static fallback
+    # endpoint; when unset the URL is discovered per-mech from the on-chain
+    # manifest (see OFFCHAIN follow-up).
```
OjusWiZard, line 240: Advisory: this and the `states/base.py` comment forward-reference behaviours that don't exist yet ("see OFFCHAIN follow-up"). Reasonable as intent docs given the PR checklist, but they're comment-rot candidates — worth a grep to refresh them when the offchain behaviours land and the naming settles.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -232,11 +232,28 @@ class MechMarketplaceConfig:
...
+    use_offchain: bool = False
+    offchain_url: Optional[str] = None
```
OjusWiZard, line 242: Advisory: `offchain_url` isn't format-validated — the guard below only checks truthiness, so `offchain_url="not-a-url"` passes. Fine for the foundation; validate the scheme in the follow-up that actually dials it (the dispatch path) rather than here.

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -42,6 +42,14 @@
 
 SERIALIZED_EMPTY_LIST = "[]"
 METADATA_FIELD = "metadata"
+
+# Off-chain dispatch `last_failure_reason` values, surfaced so operators and
+# downstream skills can branch on a stable label when the off-chain path
+# exhausts its options (see the off-chain request/response behaviours).
+OFFCHAIN_ALL_FAILED = "offchain_all_failed"
```
OjusWiZard: Advisory (non-blocking): two things on these constants.

1. **Convention split** — every existing `last_failure_reason` value is an inline string literal in the behaviours (`"subgraph_unavailable"`, `"no_non_penalized_valid_mech"`, …). These four are module-level named constants, so the codebase now has both styles, and these stay unused until the follow-up behaviours import them. Either style is fine, but consider extracting the existing reasons to constants in the same place when the offchain behaviours land, so there's one registry of failure labels.
2. **Placement** — this block is inserted mid-way through an existing run of module constants (between `METADATA_FIELD` and `BLOCK_TIMESTAMP_FIELD`), splitting that grouping. Trivial, but moving it after the existing literals reads cleaner.

OjusWiZard (dismissed):
Approving — clean, well-scoped config foundation that ships genuinely dark (new fields default off, on-chain path untouched, validation correct, four tests cover the meaningful branches). All comments below are advisory follow-ups, none blocking.

Note: I couldn't run the model tests locally (third-party AEA packages weren't synced — `ModuleNotFoundError: packages.valory.contracts.multisend`), so please re-confirm green in a synced env before merge.

OjusWiZard (dismissed):

OjusWiZard (approved):

bennyjo (approved):
Clean, correctly-scoped foundation PR — ships dark behind `use_offchain=False`, on-chain path provably unchanged. Config validation, constants, skill.yaml wiring, and tests all check out; framework FSM/resilience audits had nothing to bite on (no transitions or network calls in this diff). LGTM.

PR #94, opened 2026-06-22
feat: add MechMetadata.extra_attributes for non-prompt tool params

jmoreira-valory:
## Summary

Mech tools receive their inputs as `run(**kwargs)` — the executor spreads the request JSON's top-level keys. **mech-client** already lets callers add arbitrary top-level params via `extra_attributes` (`metadata.update(extra_attributes)`), but **`mech_interact_abci` uploaded only the fixed `MechMetadata` fields** (`asdict`), so a tool whose inputs are *structured params* rather than a natural-language `prompt` had no channel.

This adds that channel, minimally and backward-compatibly.

## Changes
- `MechMetadata`: new optional field `extra_attributes: Optional[Dict[str, Any]] = None`.
- `RequestBehaviour._send_metadata_to_ipfs`: merge `extra_attributes` into the IPFS payload **top-level** (next to `prompt`/`tool`/`nonce`) and drop the wrapper key:
  ```python
  payload = asdict(metadata)
  payload.update(payload.pop("extra_attributes", None) or {})
  ```
- 2 unit tests for the new field.

## Backward compatibility
With no `extra_attributes` set (the default), `payload.pop(...) or {}` is empty and the uploaded payload is **byte-identical** to the previous `asdict(metadata)`. Existing consumers (trader, market-resolver) are pinned to their own `mech_interact_abci` CID and are unaffected until they choose to bump.

`request_context` is **unchanged** — it remains the analysis/benchmark-context channel (uploaded for data analysis, not read as tool input). `extra_attributes` is the dedicated tool-parameter channel, mirroring mech-client.

## Motivation
The `propose-question` tool (market-creator) takes structured params (`topics`, `num_questions`, `resolution_time`) and currently has to smuggle them through a JSON-encoded `prompt`. With this change it passes them via `extra_attributes` → top-level kwargs → the tool reads them directly (no tool change).

## Test plan
- [x] `TestMechMetadata` (12 tests) + request-behaviour suite (33) pass
- [x] `autonomy packages lock` idempotent, no fingerprint pollution
- [ ] CI green

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/request.py
@@ -632,7 +632,13 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
         # prompt/tool/nonce) so the tool reads them as run() kwargs, and drop
         # the wrapper key. With no extras this is byte-identical to asdict().
         payload = asdict(metadata)
-        payload.update(payload.pop("extra_attributes", None) or {})
+        extras = payload.pop("extra_attributes", None) or {}
+        clobbered = extras.keys() & payload.keys()
+        if clobbered:
+            self.context.logger.warning(
+                f"extra_attributes override reserved request keys: {clobbered}"
+            )
+        payload.update(extras)
         metadata_hash = yield from self.send_to_ipfs(
             self.metadata_filepath, payload, filetype=SupportedFiletype.JSON
         )
--- a/packages/valory/skills/mech_interact_abci/behaviours/request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/request.py
@@ -628,8 +628,13 @@ class MechRequestBehaviour(MechInteractBaseBehaviour):
     ) -> WaitableConditionType:
         """Send Mech metadata to IPFS."""
         metadata = self._mech_requests.pop()
+        # Merge any extra tool parameters into the payload top-level (next to
+        # prompt/tool/nonce) so the tool reads them as run() kwargs, and drop
+        # the wrapper key. With no extras this is byte-identical to asdict().
+        payload = asdict(metadata)
+        payload.update(payload.pop("extra_attributes", None) or {})
         metadata_hash = yield from self.send_to_ipfs(
-            self.metadata_filepath, asdict(metadata), filetype=SupportedFiletype.JSON
+            self.metadata_filepath, payload, filetype=SupportedFiletype.JSON
         )
         if metadata_hash is None:
             return False
--- a/packages/valory/skills/mech_interact_abci/tests/states/test_base.py
+++ b/packages/valory/skills/mech_interact_abci/tests/states/test_base.py
@@ -730,6 +730,17 @@ class TestMechMetadata:
         meta = MechMetadata(prompt="q?", tool="t1", nonce="n1", request_context=ctx)
         assert meta.request_context == ctx
 
+    def test_default_extra_attributes_is_none(self) -> None:
+        """Test that extra_attributes defaults to None."""
+        meta = MechMetadata(prompt="q?", tool="t1", nonce="n1")
+        assert meta.extra_attributes is None
+
+    def test_explicit_extra_attributes(self) -> None:
+        """Test constructing with explicit extra_attributes."""
+        extra = {"topics": ["sports"], "num_questions": 1}
+        meta = MechMetadata(prompt="q?", tool="t1", nonce="n1", extra_attributes=extra)
+        assert meta.extra_attributes == extra
+
     def test_explicit_schema_version_override(self) -> None:
         """Test that schema_version can be explicitly overridden."""
         meta = MechMetadata(prompt="q?", tool="t1", nonce="n1", schema_version="3.0")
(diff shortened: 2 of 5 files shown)


packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -628,8 +628,13 @@ def _send_metadata_to_ipfs(  # pragma: no cover
     ) -> WaitableConditionType:
         """Send Mech metadata to IPFS."""
         metadata = self._mech_requests.pop()
+        # Merge any extra tool parameters into the payload top-level (next to
+        # prompt/tool/nonce) so the tool reads them as run() kwargs, and drop
+        # the wrapper key. With no extras this is byte-identical to asdict().
+        payload = asdict(metadata)
+        payload.update(payload.pop("extra_attributes", None) or {})
```
OjusWiZard: `dict.update` overwrites silently: if `extra_attributes` contains a key that collides with a reserved top-level field (`prompt`/`tool`/`nonce`/`schema_version`/`request_context`), it clobbers it with no error or log — most dangerous for the skill-generated `nonce`. This mirrors mech-client's `metadata.update(extra_attributes)`, so it's intentional parity and `extra_attributes` is consumer-controlled, but the failure mode (a corrupted on-chain request) is silent. Consider warning on collisions:

```python
extras = payload.pop("extra_attributes", None) or {}
clobbered = extras.keys() & payload.keys()
if clobbered:
    self.context.logger.warning(f"extra_attributes overrides reserved keys: {clobbered}")
payload.update(extras)
```

packages/valory/skills/mech_interact_abci/tests/states/test_base.py
```diff
@@ -730,6 +730,17 @@ def test_explicit_request_context(self) -> None:
...
+    def test_default_extra_attributes_is_none(self) -> None:
+        """Test that extra_attributes defaults to None."""
+        meta = MechMetadata(prompt="q?", tool="t1", nonce="n1")
+        assert meta.extra_attributes is None
+
+    def test_explicit_extra_attributes(self) -> None:
+        """Test constructing with explicit extra_attributes."""
+        extra = {"topics": ["sports"], "num_questions": 1}
+        meta = MechMetadata(prompt="q?", tool="t1", nonce="n1", extra_attributes=extra)
+        assert meta.extra_attributes == extra
```
OjusWiZard, line 742: These two tests cover the dataclass field's default and explicit value, but not the actual merge logic in `_send_metadata_to_ipfs` (which carries `# pragma: no cover`). The PR's three load-bearing claims are currently unverified: extras merged **top-level**, the `extra_attributes` wrapper key **dropped**, and the no-extras case **byte-identical** to the old `asdict(metadata)`. These are pure-dict assertions and don't need the IPFS round-trip:

```python
def test_payload_merges_extras_top_level_and_drops_wrapper(self) -> None:
    meta = MechMetadata(prompt="q?", tool="t1", nonce="n1",
                        extra_attributes={"topics": ["sports"], "num_questions": 1})
    payload = asdict(meta)
    payload.update(payload.pop("extra_attributes", None) or {})
    assert "extra_attributes" not in payload
    assert payload["topics"] == ["sports"] and payload["num_questions"] == 1

def test_payload_byte_identical_without_extras(self) -> None:
    meta = MechMetadata(prompt="q?", tool="t1", nonce="n1")
    payload = asdict(meta)
    payload.update(payload.pop("extra_attributes", None) or {})
    expected = {k: v for k, v in asdict(meta).items() if k != "extra_attributes"}
    assert payload == expected
```

Minor companion: `test_asdict_includes_new_fields` just below should also assert `d["extra_attributes"]`, since `asdict()` now emits it.

OjusWiZard (dismissed):
Approving. Sound, minimal, backward-compatible change — the `payload.pop(...) or {}` idiom correctly preserves byte-identity for the default path, and the field mirrors the existing `request_context` design. Two non-blocking follow-ups left inline:

- **Test coverage (important):** the merge logic in `_send_metadata_to_ipfs` is `# pragma: no cover` and untested; the byte-identity / top-level-merge claims rest on manual reasoning only.
- **Silent clobber (defensive):** `extra_attributes` keys can overwrite reserved top-level keys (`nonce`/`tool`/...) with no log. Consistent with mech-client precedent, so optional.

Neither blocks correctness on the default path.

DIvyaNautiyal07 (dismissed):

packages/valory/skills/mech_interact_abci/behaviours/request.py
jmoreira-valory: Done in 65f9713 — added the collision check + warning before the merge, exactly as suggested:
```python
extras = payload.pop("extra_attributes", None) or {}
clobbered = extras.keys() & payload.keys()
if clobbered:
    self.context.logger.warning(f"extra_attributes override reserved request keys: {clobbered}")
payload.update(extras)
```
Behaviour-only (still `# pragma: no cover`), so the silent clobber is now at least logged.

packages/valory/skills/mech_interact_abci/tests/states/test_base.py
jmoreira-valory: Done in 65f9713 — added both tests you suggested (`test_payload_merges_extras_top_level_and_drops_wrapper` + `test_payload_byte_identical_without_extras`) verbatim, verifying top-level merge, wrapper-dropped, and byte-identity without the IPFS round-trip. TestMechMetadata now 14 green.

OjusWiZard (approved):
Re-reviewed at the current head — both earlier findings are properly addressed (collision warning + merge-logic tests), byte-identity holds, and the collision check covers all reserved keys. The one residual item (tests mirror the merge expression and the warning branch is uncovered) is advisory-only and non-blocking. Approving.

DIvyaNautiyal07 (approved):

PR #95, opened 2026-06-26
feat(offchain): implement off-chain request/response behaviours

DIvyaNautiyal07:
## Summary

Builds on top of PR #93 (which landed the `use_offchain` config foundation) by shipping the behaviour-level off-chain dispatch path. With `use_offchain=True` the request and response behaviours now speak the mech server's HTTP API (`POST /send_signed_requests` and `GET /fetch_offchain_info`) instead of submitting/polling on chain. With `use_offchain=False` (the default) every existing on-chain code path runs bit-for-bit unchanged.

The implementation mirrors the wire shapes already shipped on the mech side (PRs #454, #455, #458) and on mech-client (PR #241) so the three sides stay byte-compatible.

## What's in this PR

**Off-chain request side (`behaviours/offchain_request.py`)** — `OffchainRequestExecutor`:
- Local bare-file CIDv1 ported from `mech-client`'s `compute_cidv1_bytes`. Single-block only; pinned-bytes fixture locks the algorithm.
- Client-side `request_id` derivation matching `MechMarketplace.getRequestId` (EIP-712 inner hash, contract version `"1.1.0"` hashed through `abi.encode`).
- Signing via `get_signature(digest, is_deprecated_mode=True)` so the raw 32-byte signature the contract verifies at settlement is the same shape mech-client produces.
- Form-urlencoded POST body matching the mech server's `parse_qs` parser.
- Failover across the ranked mech list (cap `offchain_failover_max_retries`, default 2) with per-mech `MechInfo.http_url` discovered from the IPFS manifest's `url` field; falls back to the static `offchain_url` config.
- Structured 402 handling: `PaymentChallenge` parsing, operator-set `auto_deposit_cap_per_cycle`, Safe-multisend deposit builder (ERC20 = `approve + depositFor` atomic; native = single payable `depositFor`; Nevermined surfaces to consumer because the BalanceTracker reverts on `depositFor` by design).
- In-flight state (`PendingRequest`) persisted on `synchronized_data.offchain_pending_request` so the deposit-then-retry round-trip re-uses the same `request_id`, signature, and on-chain nonce.

**Off-chain response side (`behaviours/offchain_response.py`)** — `OffchainResponsePoller`:
- `GET /fetch_offchain_info` polled every `offchain_poll_interval_seconds` (default 3s) with `offchain_poll_timeout_seconds` total budget (default 300s).
- Populates the existing `MechInteractionResponse` shape; consumers (`DecisionReceive` on trader, mirror downstream) read the same `mech_responses` list as the on-chain branch.

**FSM dispatch** (`states/request.py`, `rounds.py`, `states/final_states.py`):
- `MechRequestRound.end_block` reads `synchronized_data.offchain_result` and emits one of `OFFCHAIN_DONE` / `OFFCHAIN_DEPOSIT_NEEDED` / `OFFCHAIN_ALL_FAILED`, routing to three new final states with their own `db_post_conditions`. Off-path: `offchain_result=None` (the default) keeps today's `DONE` / `SKIP_REQUEST` / `BUY_SUBSCRIPTION` dispatch unchanged.
- `MechRequestBehaviour.async_act` and `MechResponseBehaviour.async_act` branch on `mech_marketplace_config.use_offchain is True` at the top. The `is True` guard (vs truthy) prevents stray mocks in tests from accidentally taking the off-chain branch.

**Manifest URL discovery** (`behaviours/mech_info.py`):
- `populate_tools` now extracts `manifest["url"]` from the same IPFS body it already fetches for `tools`, hydrating `MechInfo.http_url` for each mech in the CID group. Older manifests (no `url` key) leave the attribute as `None`, exercising the static-`offchain_url` fallback in the request behaviour.
- Operators publish the URL via `make update-metadata SERVICE=...` in `valory-xyz/mech-deployments` (see its README), which is already wired up.

**New contract wrappers**:
- `contracts/mech_marketplace` — read-only `MechMarketplace` wrapper exposing `mapNonces(sender)` and `chainId()`. The off-chain request behaviour reads both before signing each attempt so the locally-derived `request_id` matches what the contract computes at settlement. (`agent_mech` carries the same ABI but is a `.gitignore`d third-party package so we cannot extend it from this repo.)
- `contracts/balance_tracker_fixed_price_token` and `contracts/balance_tracker_fixed_price_native` — thin wrappers exposing `get_balance` (read `mapRequesterBalances`) and `build_deposit_for_data` (calldata for the 402 retry multisend). ABIs sourced from `autonolas-marketplace/abis/0.8.28/`.

**Config (`MechMarketplaceConfig`)**:
- `auto_deposit_cap_per_cycle: Optional[int]` — required when `use_offchain=True`, no default. Validator extended to enforce.
- `offchain_http_timeout_seconds: float = 60.0`
- `offchain_poll_interval_seconds: float = 3.0`
- `offchain_poll_timeout_seconds: float = 300.0`
- `offchain_failover_max_retries: int = 2`
- All five validated for sensible ranges.

## Test plan

- [x] **437 tests pass** locally (`pytest packages/valory/skills/mech_interact_abci/tests/`). 406 baseline + 31 new.
  - 23 helper tests: CIDv1 pinned-bytes fixtures (`b""`, `b"a"`, `b"hello"`), `request_id` parity per EIP-712 input field (mech, nonce, data), 402 body parsing edge cases (empty body, non-JSON, missing `payTo`, non-numeric numerics), `PendingRequest` round-trip and address lower-casing.
  - 5 round-dispatch tests: `OFFCHAIN_DONE`, `OFFCHAIN_DEPOSIT_NEEDED`, `OFFCHAIN_ALL_FAILED`, off-path `None` keeps on-chain `DONE`, unknown label falls back to `DONE` (defensive).
  - 3 manifest URL tests: `http_url` is hydrated, `http_url` stays `None` when the key is missing, `http_url` stays `None` on a malformed manifest body.
  - 8 new model validator tests: `auto_deposit_cap_per_cycle` required when `use_offchain=True`, negative cap rejected, zero cap allowed, timing/retry params validated.
- [x] `black`, `isort` formatted; `flake8` clean on touched paths.
- [x] `autonomy packages lock` refreshed; three new contract packages registered in `packages.json` with the right fingerprints.
- [ ] CI `tox -e mypy -e pylint -e darglint -e check-abciapp-specs -e check-handlers -e check-copyright -e check-doc-hashes` — not run locally; will iterate based on CI output. The FSM spec yaml may need regeneration to reflect the new transitions.

## What's deferred

- **Trader composition wiring** (one new edge `FinishedOffchainMechRequestRound → MechResponseRound` plus the multiplexer dispatch case for the 402 deposit retry). That ships as a separate PR in `valory-xyz/trader` against the new mech-interact skill hash.
- **Pre-deposit threshold + top-up behaviour** on each consumer agent (item C1 in the prepay status doc). Out of scope here.
- **Server-side signature verification** in the mech HTTP handler. Known and accepted: the contract is the arbiter at settlement, so a forged sig fails simulation and the mech only loses the LLM cost. Adding HTTP-layer recovery is operator-side DoS protection, not a correctness requirement.
- **NVM 402 auto-resolve**. By design impossible — `BalanceTrackerNvmSubscription{Native,Token}.depositFor` reverts. The path surfaces `OFFCHAIN_402_INSUFFICIENT` to the consumer.


packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+                marketplace_address=self._config.mech_marketplace_address,
+                mech_address=mech_address,
+                requester=self._safe_address(),
+                data=ipfs_data.encode("utf-8"),
+                delivery_rate=delivery_rate,
+                payment_type=payment_type_bytes,
+                nonce=on_chain_nonce,
+                chain_id=chain_id_int,
+            )
+            signature_hex = yield from self._sign_request_id(request_id_bytes)
```
OjusWiZard, line 731: **🔴 Critical — per-agent signature breaks consensus (`NO_MAJORITY` loop).**

`_sign_request_id` signs with each agent's own key, and the signature is embedded in `PendingRequest.signature` → `pending_request_json` → the `offchain_pending_request` payload field (set for both `OFFCHAIN_DONE` at L661-678 and `OFFCHAIN_DEPOSIT_NEEDED` at L693-705).

`MechRequestRound` is a `CollectSameUntilThresholdRound`; consensus is decided by `Counter(map(lambda p: p.values, payloads))` over the **full payload tuple**, which includes `offchain_pending_request`. Every agent produces a different signature → every payload tuple is unique → threshold never reached → `NO_MAJORITY` → loop back to `MechRequestRound`, which re-picks the same mech and re-signs. The off-chain happy path is **permanently wedged on any n>1 service** (single-agent works, which is likely why local tests passed).

The on-chain path is unaffected (it carries the deterministic Safe tx hash, not a per-agent sig). Fix: keep the per-agent signature out of the consensus tuple — derive it on re-entry from agreed data, move signing to a keeper round, or use the Safe/EIP-1271 signature assembled through the normal quorum path. Land it with a multi-agent round test that submits divergent `offchain_pending_request` values and asserts the round still closes.

OjusWiZard (commented):
## Multi-perspective review (6 lenses + independent consensus verification)

**Verdict: request changes.** One critical consensus bug makes the off-chain happy path non-functional on any multi-agent service, and the off-chain coordinator/poller logic ships with no behavioral test coverage (so the bug couldn't be caught).

Summary: **1 Critical** (C1 per-agent signature → `NO_MAJORITY`), **8 Important** (silent fallback, dead timeout config, silent contract-read failures, malformed-402 mis-routing, poll-loop stall, payment-hash drift, missing tests, unenforced `PendingRequest` invariants), **5 Advisory** (stringly-typed result, NVM docstring/dead enum, dead `offchain_attempted_mechs`, `Any` back-reference defeating mypy, simplifications). Details inline.

Also: the PR's own test plan flags CI `mypy`/`pylint`/`darglint`/`check-abciapp-specs` as not run locally — worth completing before merge.

_Findings personally verified against framework/code: C1 (consensus mechanism), the dead timeout config (`get_http_response` signature), and the fallback path._

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+            url = mech.http_url or self._config.offchain_url
+            if not url:
+                continue
+            return mech.address, url
+        # No ranked mech left. Fall back to the static URL if it has not
+        # already been tried — only meaningful when ``ranked_mechs`` is
+        # empty or every entry lacks a URL.
+        fallback = self._config.offchain_url
+        if fallback and not attempted:
+            return self._config.priority_mech_address or "", fallback
```
OjusWiZard: **🟠 Important — static `offchain_url` fallback silently never fires.**

When `priority_mech_address` is `None` (the normal case under dynamic selection with no eligible mech), `priority_mech_address or ""` returns `("", fallback)`. The caller's guard is `if mech_address is None or mech_url is None` — `""` passes it, so the executor appends `""` to `attempted`, queries contracts at address `""` (fails), and on the next loop `not attempted` is now False → `(None, None)` → break. The configured static URL is never POSTed to. Gate on a truthy `priority_mech_address`, or return `(None, None)` instead of `("", …)`.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -240,6 +240,28 @@ class MechMarketplaceConfig:
...
+    # token's smallest unit for ERC20). Required when ``use_offchain=True``;
+    # no default to force the operator to choose explicitly during rollout.
+    # If the 402 shortfall exceeds the cap, the behaviour refuses the
+    # deposit and surfaces ``OFFCHAIN_402_INSUFFICIENT`` to the consumer.
+    auto_deposit_cap_per_cycle: Optional[int] = None
+    # Per-POST HTTP timeout for ``/send_signed_requests``. On timeout the
+    # behaviour fails over to the next ranked mech at the same on-chain
+    # nonce. 60s default matches the prepay spec (healthy mechs comfortably
+    # beat this; the per-tool override is the configurable knob).
+    offchain_http_timeout_seconds: float = 60.0
```
OjusWiZard: **🟠 Important — `offchain_http_timeout_seconds` is dead config; comments document a timeout that doesn't exist.**

The framework `get_http_response(method, url, content, headers, parameters)` has **no timeout parameter**, and this value is referenced only in the validator — never threaded into any HTTP call. Yet this field comment, the module docstring, and `OffchainAttemptOutcome.TIMEOUT` ("The HTTP send / read timeout fired") all describe an enforced per-POST deadline. In reality `TIMEOUT` is only produced when `get_http_response` raises or returns `None`; stalls fall back to the global framework HTTP timeout. Either wire the timeout through a mechanism that supports it, or remove the config and fix the comments.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+        contract instead of trusting service config keeps the derivation
+        self-consistent across forks and replay scenarios.
+        """
+        response = yield from self._b.get_contract_api_response(
+            performative=ContractApiMessage.Performative.GET_STATE,
+            contract_address=self._config.mech_marketplace_address,
+            contract_id=str(MechMarketplace.contract_id),
+            contract_callable="get_chain_id",
+        )
+        if response.performative != ContractApiMessage.Performative.STATE:
```
OjusWiZard, line 1128: **🟠 Important — contract-read failures are silent and un-diagnosable.**

`_resolve_chain_id_int` (here), `_resolve_payment_type` (L1017), `_resolve_delivery_rate` (L1034), and the deposit builders all `return None` on a non-`STATE` performative **without logging**, then the caller maps to `OFFCHAIN_TIMEOUT_ALL_MECHS` with no further log. `_read_on_chain_nonce` (L973) already logs this correctly. An operator hitting an RPC misconfig sees `offchain_all_failed`/"timeout" with zero log lines separating chain-side from mech-side failure. Mirror the `_read_on_chain_nonce` warning in each.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+
+# ----------------------------------------------------------------------------
+# Structured 402 challenge — parses the body the mech server returns from
+# ``handlers.py::_build_402_challenge`` (mech repo). Mirrors mech-client's
+# ``PaymentChallenge`` so the cap logic and the deposit builder downstream
+# read the same field names.
+# ----------------------------------------------------------------------------
+
+
+def _safe_int(value: Any, default: int = 0) -> int:
```
OjusWiZard: **🟠 Important — malformed 402 bodies silently mis-route the deposit decision.**

`_safe_int` coerces non-integer `required`/`currentBalance` to `0`:
- malformed `required` → `shortfall=0` → `_build_deposit_tx` returns `None` (L1094) → request abandoned as `OFFCHAIN_402_INSUFFICIENT`, no log;
- malformed `currentBalance` → `shortfall=required` → spurious `OVER_CAP` (L866) that short-circuits failover even when the true shortfall was within cap;
- 402 with missing/non-string `payTo` → `parse_payment_challenge` returns `None` → `BAD_RESPONSE` → `_failure_label_for` (L1289) falls through to `OFFCHAIN_TIMEOUT_ALL_MECHS`: a chargeable 402 reported as a network timeout, no log.

Log when a default is substituted, and give `BAD_RESPONSE` its own failure label.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,300 @@
...
+
+        return urlencode({"request_id": request_id_int_str}).encode("utf-8")
+
+    def _parse(self, response: Any) -> _PollSnapshot:
+        """Translate the HTTP response into a :class:`_PollSnapshot`."""
+        if response is None:
+            return _PollSnapshot(status="processing")
+        status_code = getattr(response, "status_code", None)
+        body_bytes = getattr(response, "body", b"") or b""
+        if status_code != 200:
```
OjusWiZard, line 249: **🟠 Important — poll loop treats permanent HTTP errors as "still processing" → up to 300s stall.**

Every non-200 (404, 500, 503) maps to `status="processing"`, so a permanently-unknown `request_id` (e.g. after a mech restart) is polled for the entire `offchain_poll_timeout_seconds` budget before surfacing `poll_timeout`, and the actual status codes are never recorded at error level. Fast-fail on 404 and persistent 5xx; log distinct statuses.

_(Secondary: the per-agent wall-clock `time.monotonic()` budget in `_poll_until_terminal` is a consensus-divergence risk for `MechResponseRound` if the budget is tuned near the mech's real latency.)_

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+# hashes themselves mirror the ``PaymentType`` enum in
+# :mod:`mech_interact_abci.behaviours.request`; duplicating them here keeps
+# this module importable without pulling that file in (which would create a
+# circular import via ``MechRequestBehaviour``).
+_PAYMENT_NATIVE = "native"
+_PAYMENT_TOKEN = "token"  # nosec B105 - label string, not a credential
+_PAYMENT_NVM_NATIVE = "nvm_native"
+_PAYMENT_NVM_TOKEN = "nvm_token"  # nosec B105 - label string, not a credential
+
+_PAYMENT_TYPE_HASH_NATIVE = (
```
OjusWiZard, line 567: **🟠 Important — duplicated payment-type hashes with no drift guard.**

These five on-chain selector hashes are copy-pasted from the `PaymentType` enum in `behaviours/request.py:73-86` (the comment admits it, to avoid a circular import). A future enum change not mirrored here makes `_classify_payment_type` return `""` → the deposit path is silently skipped for that type. Extract both into a shared leaf module (e.g. `behaviours/payment_types.py`) imported by both, or at minimum add a test asserting `PaymentType.NATIVE.value == _PAYMENT_TYPE_HASH_NATIVE` etc.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -0,0 +1,363 @@
...
+#
+# ------------------------------------------------------------------------------
+
+"""Helper-level tests for the off-chain request module.
+
+Coverage focuses on the pure functions whose byte-for-byte parity with the
+mech server is the single highest-risk silent-failure mode: local CIDv1
+computation, ``request_id`` derivation, structured 402 parsing, and the
+in-flight state round-trip. Behaviour-level integration tests cover the
+executor's failover decision tree.
```
OjusWiZard, line 26: **🟠 Important — off-chain coordinator/poller/deposit builders have no behavioral tests, and this docstring is false.**

This docstring claims integration tests cover the executor's failover decision tree — none exist. The 31 new tests cover only the pure-helper layer. **Untested:** `OffchainRequestExecutor.run/_fresh_cycle/_retry_pending`, the failover loop, `OffchainResponsePoller.run/_poll_until_terminal`, `_classify_payment_type`, all deposit-tx builders, and `_payload_from_offchain_result`. `async_act` is `# pragma: no cover`, which also drags `_run_offchain_*_cycle` out of coverage. This is why the Critical consensus bug shipped. Add: a `_fresh_cycle` test with mocked reads/HTTP, a `_poll_until_terminal` timeout test, and a multi-agent consensus test for the payload.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+    ipfs_data: str
+    signature: str  # hex with 0x prefix
+    attempted_mechs: List[str] = field(default_factory=list)
+
+    def to_json(self) -> str:
+        """Serialize to the form persisted on synchronized_data."""
+        return json.dumps(dataclasses.asdict(self))
+
+    @classmethod
+    def from_dict(cls, raw: Dict[str, Any]) -> Optional["PendingRequest"]:
```
OjusWiZard, line 523: **🟠 Important — `PendingRequest` format invariants unenforced; deposit-retry can crash.**

`request_id` must be hex **without** `0x` and `signature` hex **with** `0x` (comment-only). `from_dict` does `str(raw[...])` with no normalization, and `_retry_pending` calls `bytes.fromhex(pending.request_id)` (L749), which raises `ValueError` on a `0x`-prefixed value. A `__post_init__` assertion (allowed on frozen dataclasses) would fail fast at construction instead of mid-retry.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+_PAYMENT_TYPE_HASH_NVM_TOKEN = (  # nosec B105 - public on-chain payment-type selector hash
+    "0x0d6fd99afa9c4c580fab5e341922c2a5c4b61d880da60506193d7bf88944dd14"
+)
+
+
+@dataclass(frozen=True)
+class OffchainCycleResult:
+    """Final outcome of one offchain request cycle, lifted onto the payload."""
+
+    offchain_result: (
```
OjusWiZard: **🟡 Advisory — stringly-typed `offchain_result`.**

This `str` (9 literal sites) must stay in lockstep with `_OFFCHAIN_RESULT_TO_EVENT` (`states/request.py`) and the `Event` enum. A typo makes the dict `.get` return `None` and the round silently falls through to the on-chain `DONE`/`SKIP_REQUEST` branch — no error. Use a `Literal[...]`/enum, or key the dict off `Event.*.value` so a rename fails type-checking.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+                        sender=self._safe_address(),
+                        delivery_rate=delivery_rate,
+                        ipfs_hash=ipfs_hash,
+                        ipfs_data=ipfs_data,
+                        signature=signature_hex,
+                        attempted_mechs=attempted,
+                    ).to_json(),
+                )
+
+            if attempt.outcome in (
```
OjusWiZard: **🟡 Advisory — NVM docstring vs behavior; dead enum member.**

`_build_deposit_tx` says the NVM path "short-circuit[s] failover," but for an NVM 402 the outcome is `DEPOSIT_NEEDED` → `_build_deposit_tx` returns `None` → the caller `continue`s and burns the full failover budget before failing. Also `OffchainAttemptOutcome.NVM_UNSUPPORTED` is **never produced**, so this `(OVER_CAP, NVM_UNSUPPORTED)` branch only ever fires for `OVER_CAP`. Net result lands on `OFFCHAIN_402_INSUFFICIENT` either way, but the code/docs mislead.

packages/valory/skills/mech_interact_abci/payloads.py
```diff
@@ -42,6 +42,18 @@ class MechRequestPayload(PrepareTxPayload):
...
+    # Offchain dispatch fields. None for the on-chain path (today's behaviour,
+    # unchanged). One of ``offchain_done|offchain_deposit_needed|
+    # offchain_all_failed`` when ``MechRequestBehaviour`` took the offchain
+    # HTTP path. ``MechRequestRound.end_block`` reads ``offchain_result`` to
+    # dispatch to the right ``Event`` variant; the other three carry the
+    # state needed to either retry after a settled deposit, continue failover
+    # across the ranked mech list, or surface a clean failure to the
+    # consumer.
+    offchain_result: Optional[str] = None
+    offchain_attempted_mechs: Optional[str] = None
```
OjusWiZard: **🟡 Advisory — `offchain_attempted_mechs` is dead plumbing.**

This payload field is always set to `None` (`request.py:1046`), so `SynchronizedData.offchain_attempted_mechs` always returns `[]`; the real failover history lives in `PendingRequest.attempted_mechs`. It's also in `MechRequestRound.selection_key`, so removing it is a protocol change needing a coordinated bump — but it is genuinely dead and worth removing.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+__all__ = [
+    "OffchainResponsePoller",
+    "serialise_responses",
+]
+
+
+# Silence unused-import warnings for symbols only used in type annotations.
+_ = Dict
```
OjusWiZard: **🟡 Advisory — dead import + wrong comment (representative of several simplifications).**

`Dict` is not used in any annotation in this file, so the comment ("symbols only used in type annotations") is wrong and `_ = Dict` just masks a dead import — remove both. Other behavior-preserving cleanups flagged in review: dedupe the two identical `PendingRequest(...)` blocks (offchain_request.py:660-705); extract the repeated `performative != STATE → return None` guard (8 sites); hoist local `import hashlib` / `from urllib.parse import urlencode` to module scope; drop the single-use `_attempt_single_mech` pass-through; simplify the unreachable `str` branch in `OffchainResponsePoller._load_pending`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. Dropped the signature field from PendingRequest, so each agent now re-signs request_id_bytes locally inside the executor on both the fresh-cycle POST and _retry_pending. The remaining fields are deterministic given the on-chain reads, so the round's CollectSameUntilThresholdRound consensus over the payload tuple converges across agents. Added a multi-agent test in test_rounds.py (test_consensus_with_deterministic_pending_request) that sends the same pending blob from every participant and asserts the round emits OFFCHAIN_DONE, plus a no_signature_field invariant test on PendingRequest so a future refactor cannot quietly reintroduce divergence.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. _pick_next_mech now requires a truthy priority_mech_address before returning the static fallback, and also skips it when the priority address is already in the attempted set. Added behavioural tests in TestPickNextMechFallback covering all three branches: no priority address yields (None, None), a real priority + static URL is returned correctly, and a previously-attempted static priority is skipped.

packages/valory/skills/mech_interact_abci/models.py
DIvyaNautiyal07: Fixed in e5771f7. Removed offchain_http_timeout_seconds from MechMarketplaceConfig, dropped its validator branch, and removed it from skill.yaml and test_models.py. The OffchainAttemptOutcome.TIMEOUT docstring now states what the outcome actually means: the HTTP call raised at the framework boundary or returned None, not a locally-policed deadline. The module-level docstring still references failover on timeout / 503, which is accurate for the framework's own timeout signal.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. Added warning logs in _resolve_chain_id_int, _resolve_payment_type, _resolve_delivery_rate, _build_native_deposit_tx, and _build_token_deposit_multisend on both the non-STATE performative branch and the malformed-body branch. Each log line names the contract call and the failure shape (performative or value type), so an operator looking at the agent log can immediately tell chain-side RPC misconfig apart from mech-side failure.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. Replaced _safe_int (which silently coerced to 0) with _try_int that returns Optional[int]. parse_payment_challenge now treats a malformed required or currentBalance as a 'not a structured 402' and returns None, which the caller maps to OffchainAttemptOutcome.BAD_RESPONSE. Added a new OFFCHAIN_BAD_RESPONSE label so the failover loop surfaces it distinctly instead of falling through to OFFCHAIN_TIMEOUT_ALL_MECHS. chainId stays tolerant of missing values (defaults to 0) because the deposit builder doesn't consult it. Tests cover non-numeric required, non-numeric currentBalance, boolean required, and missing chainId.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in e5771f7. The poll loop now fast-fails on 404 (rejected/not_found) and on 5 consecutive 5xx responses (rejected/server_unavailable). The 5xx counter resets on any non-5xx status, so a flaky mech that briefly recovers doesn't trip the fast-fail. Non-200 logging is now owned by the caller so each fast-fail branch can log at the appropriate level (warning for the counter, error for the terminal). Added a behavioural test file test_offchain_response_poller.py with 404 fast-fail, 5xx fast-fail, counter-reset, ok/rejected dispatch, and poll-timeout cases. The wall-clock vs counter consensus-divergence point is real but is a separate fix tied to MechResponseRound restructuring; happy to file a follow-up if useful.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. Added a parametrised drift-guard test TestPaymentTypeHashesMatchEnum::test_payment_type_hash_matches_enum that asserts each of the five _PAYMENT_TYPE_HASH_* constants equals the matching PaymentType.*.value. A future enum change without a matching constant update will fail this test in CI before the deposit dispatch can silently skip. Went with the test over extracting a shared leaf module so the no-circular-import comment in the executor stays honest.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Fixed in e5771f7. The docstring no longer claims behavioural coverage that doesn't exist. Added 32 behavioural tests in two files: test_offchain_request_helpers.py now has TestPickNextMechFallback, TestFreshCycle (DONE on first try, 402-under-cap deposit-needed, 402-over-cap short-circuit, timeout-then-DONE failover, all-timeouts-all-failed, chain-id read failure), TestRetryPending (retry-DONE, second-402), TestClassifyPaymentType, TestAttemptResultPlumbing (status → outcome mapping, malformed-402 → BAD_RESPONSE), plus the C7 drift guard. The new test_offchain_response_poller.py covers _poll_until_terminal (404 fast-fail, 5 consecutive 5xx, counter-reset, ok/rejected, poll-timeout). test_rounds.py has a multi-agent consensus test for the payload. 481 tests pass (75 new on top of the original 406).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. PendingRequest now has a __post_init__ that asserts request_id matches ^[0-9a-f]{64}$ (no 0x prefix), mech_address equals its lowercase form, and nonce/delivery_rate are non-negative. from_dict normalises both leading-0x and uppercase request_ids to the canonical form so a re-entry against a payload written by an older revision still validates. Added tests covering malformed request_id, uppercase mech_address, negative nonce, and the 0x-prefix normalisation path. The signature field invariant is moot now that the field itself is gone (C1).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. _OFFCHAIN_RESULT_TO_EVENT is now keyed off Event.OFFCHAIN_*.value, and every emit site in offchain_request.py uses Event.OFFCHAIN_*.value rather than free string literals. A future rename of the event symbol or its value will break the dispatch dict at construction time instead of silently routing to the on-chain DONE branch. OffchainCycleResult.offchain_result still types as str (the round dispatches on the value, not the enum), but its docstring now spells out the allowed values come from Event.OFFCHAIN_*.value.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e5771f7. Rewrote the _build_deposit_tx docstring to match actual behaviour: NVM and unknown payment types return None, and the caller treats any None as OFFCHAIN_402_INSUFFICIENT (not as failover) because the other ranked mechs would charge the same Safe balance and reach the same outcome. Removed OffchainAttemptOutcome.NVM_UNSUPPORTED (never produced) and simplified the OVER_CAP/NVM branch to just check OVER_CAP. Added a warning log on unknown paymentType selector so a misconfigured chain doesn't fail silently.

packages/valory/skills/mech_interact_abci/payloads.py
DIvyaNautiyal07: Fixed in e5771f7. Removed offchain_attempted_mechs from MechRequestPayload, SynchronizedData, MechRequestRound.selection_key, the executor (it was reading the synced value to seed the local attempted list), the offchain payload builder, and skill.yaml. The real failover history lives in the executor's local attempted list for the duration of one cycle; it does not need to cross rounds since the cycle either completes (DONE/DEPOSIT_NEEDED/ALL_FAILED) or returns to a fresh _fresh_cycle. Caught this is technically a protocol change pre-merge, but since use_offchain ships dark (default False), the on-chain branch is unaffected.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in e5771f7. Dropped the _ = Dict line and the unused Dict import in offchain_response.py. Hoisted hashlib (compute_cidv1_bytes) and urlencode (_build_form_body, _build_body) to module scope. Consolidated the two identical PendingRequest(...) constructor blocks for DONE/DEPOSIT_NEEDED into a single _build_pending helper so the field shape can't drift between the two branches. Dropped the _attempt_single_mech pass-through (the caller now calls _post_signed_request directly). Simplified _load_pending in the poller: SynchronizedData.offchain_pending_request already normalises the JSON-string and dict shapes, so the unreachable str branch is gone.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1308 @@
...
+    def _build_native_deposit_tx(
+        self, challenge: PaymentChallenge
+    ) -> Generator[None, None, Optional[str]]:
+        """Single Safe tx: ``BalanceTracker.depositFor(safe)`` with ``value=shortfall``."""
+        # Encode the depositFor selector — value is supplied by the Safe
+        # tx envelope, not by the calldata, mirroring mech-client's native
+        # ``deposit_native`` path.
+        call_data_response = yield from self._b.get_contract_api_response(
+            performative=ContractApiMessage.Performative.GET_STATE,
+            contract_address=challenge.pay_to,
```
bennyjo, line 1480: **Impact:** `pay_to` and `asset` come straight from the mech's 402 body, unvalidated. The native path sends `value=shortfall` to that address; the token path does `approve(pay_to, shortfall)` on a mech-chosen `asset`. So a mech can direct a real transfer/approval — up to `auto_deposit_cap_per_cycle`, every cycle — to an address of its choosing, instead of a BalanceTracker that credits your Safe.

**Likelihood:** needs a malicious or compromised mech, or a MITM on the endpoint (nothing forces `https` on the manifest `url`). Bounded by the cap, but recurring and to an arbitrary recipient.

Can we assert `challenge.pay_to` equals the BalanceTracker bound to this mech on-chain (and `asset` its token) before depositing? You already read `paymentType`/`maxDeliveryRate` from the mech, so the tracker is reachable. Today the cap is the only guard.

bennyjo (commented):
One blocking finding on the 402 deposit path (attacker-controlled deposit target). Other items are in the review writeup as follow-ups.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 1bf0c10. Good catch, thanks. Before _build_deposit_tx, the executor now reads MechMarketplace.mapPaymentTypeBalanceTrackers(paymentType) and asserts challenge.pay_to matches. On the token path it also reads BalanceTrackerFixedPriceToken.token() and asserts challenge.asset matches. A mismatch (or a tracker read failure, or a zero-address registration) is treated as BAD_RESPONSE so failover continues to the next ranked mech instead of landing a redirected deposit. Added MechMarketplace.get_balance_tracker and BalanceTrackerFixedPriceToken.get_token contract wrappers. Tests cover pay_to mismatch, asset mismatch on the token path, matching destination proceeds to build, tracker read failure, and zero-address tracker registration. The cap stays in place as defence in depth.

packages/valory/skills/mech_interact_abci/rounds.py
```diff
@@ -164,6 +179,9 @@ class MechInteractAbciApp(AbciApp[Event]):
         FinishedMechResponseTimeoutRound,
         FinishedMechRequestSkipRound,
         FinishedMechPurchaseSubscriptionRound,
+        FinishedOffchainMechRequestRound,
+        FinishedOffchainMechDepositNeededRound,
+        FailedOffchainMechRequestRound,
     }
     event_to_timeout: EventToTimeout = {
         Event.ROUND_TIMEOUT: 30.0,
```
OjusWiZard, line 193: **🟠 Important — the 300s offchain poll budget is unreachable; this 30s round timeout is the real ceiling.**

`MechResponseRound` shares this single `ROUND_TIMEOUT: 30.0` (and `round_timeout_seconds: 30.0` in skill.yaml). I traced the framework: `AbciApp.update_time` fires `process_event(ROUND_TIMEOUT)` as soon as block time passes `last_timestamp + 30s` → transitions to `FinishedMechResponseTimeoutRound`; on the next tick `BaseBehaviour.act()` sees the round changed and abandons `OffchainResponsePoller._poll_until_terminal` mid-loop, so no payload is ever submitted.

Consequences:
- Any mech that takes more than ~30s to deliver **always** ends in `FinishedMechResponseTimeoutRound`, regardless of `offchain_poll_timeout_seconds`.
- The deadline check in `_poll_until_terminal` (`if time.monotonic() >= deadline`) and the `offchain_poll_timeout_seconds: 300.0` default in skill.yaml are effectively inert on this path — setting it to 60 or 240 changes nothing.
- This bites off-chain specifically: it polls a live LLM (commonly 10–120s). The on-chain path shares the structure but settles in seconds, so it never trips.

**Fix:** make the two coherent — give the offchain response round a dedicated timeout ≥ `offchain_poll_timeout_seconds` + overhead, or define `offchain_poll_timeout_seconds` as advisory and document that `round_timeout_seconds` is the true bound.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,337 @@
...
+            )
+            return _PollSnapshot(status="processing")
+        if not isinstance(payload, dict) or not payload:
+            return _PollSnapshot(status="processing")
+        status = str(payload.get("status", "")).lower()
+        if status == "ok":
+            return _PollSnapshot(
+                status="ok",
+                result=self._serialise_result(payload.get("response")),
+                content_cid=payload.get("content_cid"),
```
OjusWiZard: **🟡 Advisory — `content_cid` is parsed here but never consumed.**

`_PollSnapshot.content_cid` (declared L68) is populated on the `status="ok"` branch, but `_apply_snapshot` only reads `result`/`error`, and `MechInteractionResponse` has no field to hold it — so it's discarded on every successful response. Either drop it from `_PollSnapshot`, or thread it into `MechInteractionResponse` if a consumer is meant to use it for result/IPFS verification.

OjusWiZard (commented):
## Second-pass review (current head `1bf0c10`)

Round-one findings are all fixed/resolved, and the deposit-redirection issue (bennyjo) is properly closed by validating `payTo`/`asset` against the on-chain `mapPaymentTypeBalanceTrackers` + `BalanceTracker.token()`. 👍

Two residual items, no blockers: **1 Important** (300s poll budget vs 30s response-round timeout — independently verified in the framework), **1 Advisory** (dead `content_cid`). Details inline.

_Cleared below the confidence bar: the `1bf0c10` 402-destination validation is correct (native-skips-asset, `""`-label short-circuit, lowercase compares all fine); the multi-agent per-signature POST is intentional and its only open question (mech-side request-id dedup / double-charge) lives in the mech server, not this repo — worth a one-line confirm with that team._

packages/valory/skills/mech_interact_abci/rounds.py
DIvyaNautiyal07: Fixed in d0a929a. Good catch, the 30s timeout was the real ceiling. Added a dedicated Event.RESPONSE_ROUND_TIMEOUT and routed MechResponseRound's timeout transition through it instead of the shared ROUND_TIMEOUT. The class-level event_to_timeout carries a 330s fallback (300s default poll budget + 30s overhead). SharedState.setup rebinds it at runtime from mech_marketplace_config.offchain_poll_timeout_seconds so an operator-tuned poll budget stays coherent with the round timeout that bounds it. Other rounds keep the 30s ROUND_TIMEOUT. Updated fsm_specification.yaml and added tests that pin the rebind contract and that the class-level default covers the 300s default poll budget.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in d0a929a. Dropped content_cid from _PollSnapshot and stopped populating it in _parse. No downstream consumer was reading it. If we end up needing it for IPFS verification later, the place to add it is MechInteractionResponse so the field actually crosses round boundaries.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1624 @@
...
+def _compute_domain_separator(chain_id: int, marketplace_address: str) -> bytes:
+    """Reproduce ``MechMarketplace._computeDomainSeparator`` in Python.
+
+    The contract hashes the version through ``abi.encode`` not as raw
+    bytes (see ``MechMarketplace.sol:160``), so the Python side must
+    encode-then-hash the version string too — a raw ``keccak256(b"1.1.0")``
+    would not match.
+    """
+    name_hash = eth_keccak(text=_MARKETPLACE_NAME)
+    version_hash = eth_keccak(abi_encode(["string"], [_MARKETPLACE_VERSION]))
```
OjusWiZard, line 226: **Verify against the deployed contract before this hits production.** The EIP-712 domain separator hashes `version` via `abi.encode` (line 223) but `name` as raw UTF-8 bytes (line 222). Standard EIP-712 / OZ `EIP712.sol` hash both as `keccak256(bytes(s))`, so this asymmetry is unusual. The docstring cites `MechMarketplace.sol:160`, but nothing pins it: the parity tests in `test_offchain_request_helpers.py` only assert length-32 and input-divergence — there is no golden vector (the CIDv1 path has pinned-byte fixtures; this path does not).

If the contract actually follows standard EIP-712 for `version`, every off-chain `request_id` fails signature verification at settlement **silently** (rejected settlement, no exception). Please add a golden-value test asserting `derive_request_id_bytes(...)` equals a `request_id` produced by the real `MechMarketplace.getRequestId` (or mech-client) so this is locked forever.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1624 @@
...
+                if tx_hex is None:
+                    # Either NVM (not on-chain auto-resolvable) or the
+                    # deposit builder hit a contract-read failure; both
+                    # are non-recoverable on this cycle, so surface to
+                    # the consumer instead of burning failover budget.
+                    return OffchainCycleResult(
+                        offchain_result=Event.OFFCHAIN_ALL_FAILED.value,
+                        last_failure_reason=OFFCHAIN_402_INSUFFICIENT,
+                    )
+                return OffchainCycleResult(
```
OjusWiZard, line 806: **Data loss on the deposit-retry success path.** This `DEPOSIT_NEEDED` result omits `mech_requests_json` (→ `None`). `mech_requests` is in `MechRequestRound.selection_key` (`states/request.py:64`), so the round writes `null`, and the accessor collapses it to `[]` (`states/base.py:455`, `requests or []`). When `_retry_pending` later succeeds it reads `self._synced.mech_requests` (now `[]`, line 863-864) and the consumer at `FinishedOffchainMechRequestRound` loses the original request metadata.

The fresh-cycle DONE path (line 759) deliberately carries `[request_meta]` on the result — the deposit branch should too:
```python
mech_requests_json=self._serialise_mech_requests([request_meta]),
```
Latent until the deposit-retry edge is wired (the deferred trader PR), but the buggy code ships here. A regression test driving DEPOSIT_NEEDED → retry-DONE and asserting `mech_requests` survives would have caught this.

OjusWiZard (dismissed):
Approving. Two items worth addressing before this carries production traffic — left inline. Neither blocks the on-chain default path (`use_offchain=False` runs unchanged); both live on the off-chain branch.

bennyjo (dismissed):
Blocking finding resolved: the 402 deposit destination is now validated against the marketplace's on-chain BalanceTracker registry (`mapPaymentTypeBalanceTrackers`), with `asset` checked against `token()` on the ERC20 path, failing closed on mismatch/read-failure/zero-address, and covered by discriminating tests. Follow-up items (poll-budget vs round timeout, signature-in-consensus, 402 parsing, fast-fail) also addressed. Approving.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in d332d04. I verified against the source at MechMarketplace.sol:155-165 and 891-907 — the asymmetric domain encoding (name as raw keccak, version through abi.encode) is genuinely what the contract does, so the Python mirror is correct. But you're right that the existing tests would not catch a silent drift if someone 'fixed' the version hashing to match standard EIP-712. Added test_golden_vector_matches_eip712_primitives in TestDeriveRequestIdBytes that reconstructs the expected request_id step-by-step using eth_abi.encode and eth_utils.keccak directly, mirroring both Solidity sources line for line, then pins the intermediate domain separator (df50cdb…) and the final request_id (76a226b…). Any drift in either layer fails the test.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in d332d04. Good catch. The DEPOSIT_NEEDED branch now serialises [request_meta] into mech_requests_json just like the DONE path, so the round commits real metadata rather than null and the retry leg's read of self._synced.mech_requests preserves prompt/tool/nonce. Added test_deposit_needed_then_retry_preserves_mech_requests that drives the full round trip (fresh cycle returns DEPOSIT_NEEDED with metadata, simulate the round commit, second leg runs _retry_pending with the rehydrated synced data and returns OFFCHAIN_DONE) and asserts the original prompt/tool/nonce values survive to the final mech_requests_json. Extended test_402_under_cap_returns_deposit_needed to assert the metadata presence on the DEPOSIT_NEEDED result directly.

OjusWiZard (dismissed):

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1700 @@
...
+        )
+        if response.performative != ContractApiMessage.Performative.STATE:
+            self._logger.warning(
+                f"Mech {mech_address} maxDeliveryRate read failed: "
+                f"performative={response.performative}"
+            )
+            return None
+        raw = response.state.body.get("max_delivery_rate")
+        try:
+            return int(raw)
```
OjusWiZard: **A zero/invalid `delivery_rate` silently defeats dynamic sizing.** `max_delivery_rate` is returned as-is here, including `0`. Since `0 is not None`, the `if delivery_rate is None: continue` guard in `_fresh_cycle` (L716-719) doesn't catch it, so `_compute_deposit_amount` computes `desired = target_calls × 0 = 0` and `needed = max(shortfall, -current_balance) = shortfall`. The deposit silently collapses back to one-call-at-a-time sizing — exactly the behavior this commit removes — and every subsequent request 402s again, with no log.

Suggest guarding non-positive rates as an invalid read so failover skips the mech:
```python
rate = int(raw)
if rate <= 0:
    self._logger.warning(f"Mech {mech_address} maxDeliveryRate is {rate}; skipping.")
    return None
return rate
```

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1700 @@
...
+                # ``prompt``/``tool``/``nonce`` as empty.
+                return OffchainCycleResult(
+                    offchain_result=Event.OFFCHAIN_DEPOSIT_NEEDED.value,
+                    tx_submitter=OFFCHAIN_DEPOSIT_TX_SUBMITTER,
+                    tx_hash=tx_hex,
+                    mech_requests_json=self._serialise_mech_requests([request_meta]),
+                    pending_request_json=pending.to_json(),
+                )
+
+            if attempt.outcome is OffchainAttemptOutcome.OVER_CAP:
```
OjusWiZard, line 814: **This OVER_CAP path — the one that actually fires — logs nothing, while the only "cap too low" warning sits in dead code.** OVER_CAP is decided upstream at L985-986 (`if challenge.shortfall > cap`), so by the time `_build_deposit_tx`/`_compute_deposit_amount` runs, `shortfall <= cap` is guaranteed. That makes the `if cap and shortfall > cap: return None` guard at L1453 **and** its warning at L1404-1410 unreachable — they can never execute. Meanwhile this branch surfaces `OFFCHAIN_402_INSUFFICIENT` with no log, indistinguishable from the NVM and unknown-payment-type paths that surface the same string.

Suggest moving the warning here (it has `attempt.challenge`), and dropping the dead guard at L1453 (and its docstring paragraph):
```python
if attempt.outcome is OffchainAttemptOutcome.OVER_CAP:
    self._logger.warning(
        f"402 shortfall {attempt.challenge.shortfall} exceeds "
        f"auto_deposit_cap_per_cycle {self._config.auto_deposit_cap_per_cycle}; "
        "cap too low for this mech. Surfacing OFFCHAIN_402_INSUFFICIENT."
    )
    return OffchainCycleResult(...)
```

packages/valory/skills/mech_interact_abci/states/request.py
```diff
@@ -33,12 +33,38 @@
...
 
+# Sentinel ``tx_submitter`` the off-chain deposit retry path stamps onto
+# ``OffchainCycleResult.tx_submitter``. Consumer multiplexers (e.g.
+# ``liquidity_trader_abci.PostTxSettlementRound``) key off this value to
+# route the settled deposit tx back into ``MechRequestRound`` — where
+# ``_retry_pending`` runs the cached POST — instead of treating it like a
+# normal mech request tx (which dispatches forward to ``MechResponseRound``).
+# The value is intentionally distinct from ``MechRequestRound.auto_round_id()``
+# (``"mech_request_round"``) so the multiplexer can tell the two cases apart.
+OFFCHAIN_DEPOSIT_TX_SUBMITTER = "mech_request_round_offchain_deposit"
```
OjusWiZard, line 44: **The sentinel value isn't pinned by any test — a rename breaks consumers silently.** `test_402_under_cap_returns_deposit_needed` asserts `result.tx_submitter == OFFCHAIN_DEPOSIT_TX_SUBMITTER`, importing the same symbol on both sides — an identity check, not a value pin. External consumers (the trader/optimus multiplexer) hardcode the literal `"mech_request_round_offchain_deposit"` to route the settled deposit back into MechRequestRound; if this constant is renamed, both sides of that assertion move together and the test stays green while the consumer routing breaks. Add a literal pin:
```python
assert OFFCHAIN_DEPOSIT_TX_SUBMITTER == "mech_request_round_offchain_deposit"
```
(The old broken literal `"MechRequestRound"` is fully gone from functional paths — confirmed.)

OjusWiZard (commented):
Follow-up pass on the new dynamic-deposit-sizing + sentinel-tx_submitter logic (commit 68e974e). The sizing arithmetic is provably sound (deposit always ≥ shortfall, ≤ cap) and the golden-vector test is a genuine independent pin — nice. Four items inline: two silent-failure/diagnosability gaps and two test-hardening asks. No Critical/security issues in the new code.

(Note: I also checked the `cap == 0` semantics — the upstream gate at L985-986 makes `cap=0` refuse every real 402, i.e. *disabled*, so the existing test docstring is correct and there's nothing to fix there.)

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -254,6 +284,27 @@ def __post_init__(self) -> None:
...
+                "use_offchain requires auto_deposit_cap_per_cycle to be set "
+                "(operator-required cap on a single 402-triggered deposit, "
+                "in the payment asset's smallest denomination)"
+            )
+        if (
+            self.auto_deposit_cap_per_cycle is not None
+            and self.auto_deposit_cap_per_cycle < 0
+        ):
+            raise ValueError("auto_deposit_cap_per_cycle must be non-negative")
+        if self.offchain_deposit_target_calls < 1:
```
OjusWiZard, line 298: **This `offchain_deposit_target_calls` validator has no test coverage.** The parametrized `test_invalid_offchain_timing_params_raise` covers the poll/retry params but not this one — neither the `>= 1` rejection (`0`/`-1`) nor the default of `10` is asserted anywhere. It's the entry guard for the sizing formula, so a silent drop during a future `__post_init__` refactor would let nonsensical config through unnoticed. Worth one parametrized case here plus a default-value assertion in `test_models.py`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in b955963. _resolve_delivery_rate now rejects any rate <= 0 with a warning so failover skips the mech. Added two regression tests in TestFreshCycle (zero rate, negative rate) that drive _fresh_cycle and assert the failure routes through OFFCHAIN_TIMEOUT_ALL_MECHS.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in b955963. Moved the warning to the OVER_CAP branch in _fresh_cycle (where the outcome actually emits, with access to attempt.challenge.shortfall and the cap). Dropped the dead guard in _compute_deposit_amount plus the corresponding deposit_amount=None call-site check and the test that pinned it. Narrowed the return type to int. Reachability invariant documented in the docstring: _post_signed_request filters shortfall > cap upstream so by the time _compute_deposit_amount runs, shortfall <= cap is guaranteed and the clamp always satisfies the current request.

packages/valory/skills/mech_interact_abci/states/request.py
DIvyaNautiyal07: Fixed in b955963. Added a literal pin alongside the existing identity check: assert result.tx_submitter == 'mech_request_round_offchain_deposit'. A rename of the constant now fails this assertion immediately.

packages/valory/skills/mech_interact_abci/models.py
DIvyaNautiyal07: Fixed in b955963. Extended the parametrize in test_invalid_offchain_timing_params_raise with offchain_deposit_target_calls = 0 and -1 cases (both expect 'must be >= 1'), and added test_offchain_deposit_target_calls_default that asserts the default of 10.

OjusWiZard (approved):
Approving. The off-chain request/response implementation has converged: all earlier review findings (Critical through Advisory) are addressed, all 22 review threads are resolved, and the latest fixes (delivery-rate guard, OVER_CAP diagnosability, sentinel value pin, validator coverage) landed cleanly in b955963.

Verified at this state:
- Deposit-sizing arithmetic is provably sound (deposit always ≥ shortfall, ≤ cap).
- `request_id` EIP-712 derivation is pinned by an independent golden vector.
- `mech_requests` is carried across the DEPOSIT_NEEDED → retry path (no data loss).
- CI fully green: linter, lock_check, gitleaks, copyright, and tests across macOS/Ubuntu/Windows × Python 3.10–3.14.
- mergeStateStatus CLEAN, mergeable.

Deferred items (OffchainCycleResult discriminated-union refactor, darglint docstring polish) are non-blocking and tracked for follow-up. LGTM. 🚀

bennyjo (approved):
Re-reviewed the delta since my last approval (`d0a929a`): dynamic deposit sizing + sentinel `tx_submitter`, and `mech_requests` now carried on the `DEPOSIT_NEEDED` branch.

The new sizing helper is sound and bounded — `min(max(shortfall, target_calls·delivery_rate − current_balance), cap)`, with the reachability invariant (`shortfall ≤ cap` upstream) guaranteeing the current request's deposit is always covered; attacker-influenced `current_balance` stays bounded by the cap and the destination is still validated against the on-chain BalanceTracker registry. Zero/negative `delivery_rate` now correctly rejected. The `mech_requests_json` carry on `DEPOSIT_NEEDED` fixes the empty-metadata-on-retry bug. Test coverage is discriminating (exact-value sizing cases, full DEPOSIT_NEEDED→retry round trip, request_id golden vector). CI green across the matrix.

Non-blocking note: `auto_deposit_cap_per_cycle` semantics shifted — the deposit now routinely sizes up to the full cap (default 10× a single call) rather than just the shortfall, so already-deployed services should revisit their configured cap. Documented in the config comment; bounded and credited to the Safe's own prepaid balance.

Approving.

PR #3, opened 2025-08-26
Purchase NVM subscription

KahanMajmudar:
## Proposed changes

WIP: Adds nvm purchase workflow

## Fixes

If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [x] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #2, opened 2025-08-22
Implement NVM functionality

Adamantios:

PR #96, opened 2026-06-30
fix(offchain): 3 executor bugs from #95 uncovered by local e2e

DIvyaNautiyal07:
## Summary

Three bugs in the off-chain executor introduced by #95 that escaped CI because no test exercised the executor end-to-end against a live mech. All surfaced in the first local e2e run against an Anvil Base fork.

1. **chain_id routing** — every `get_contract_api_response` call in `OffchainRequestExecutor` was missing `chain_id=self._b.params.mech_chain_id`. The contract dispatcher falls back to the AEA ledger id (`"ethereum"`), which has no entry in multi-chain consumers' `ledger_apis`. The lookup returns empty and `EthereumApi` constructs against its default mainnet RPC — where the marketplace at the configured address is not deployed, so every read reverts with `performative=error`. The on-chain path (`base.py`, `request.py`) already passes `chain_id` on every call; this restores parity.

2. **`_pick_next_mech` not gated on `use_dynamic_mech_selection`** — `ranked_mechs` cascades to `self.mech_tool` (`db.get_strict`), which is only written by the v2 dynamic-selection round. Consumers using a hardcoded `priority_mech_address` never set it, so every offchain cycle raised `'mech_tool' field is not set for this period`. The on-chain selector in `request.py:574-585` already wraps the same traversal in `if use_dynamic_mech_selection:`; this mirrors that gate. The static-URL fallback below stays intact.

3. **`_build_safe_tx_for_single_call` referenced a nonexistent helper** — it called `self._b._get_safe_tx_hash(...)`, but `MechRequestBehaviour` only has a string attribute named `_safe_tx_hash`. Every deposit-needed cycle crashed with `AttributeError`. Replaced with the inline `GnosisSafeContract.get_raw_safe_transaction_hash` call that `MechRequestBehaviour._build_multisend_safe_tx_hash` uses in `base.py`; only the operation differs (CALL vs DELEGATE_CALL).

## Test plan

- [ ] CI green
- [ ] Local e2e against Anvil Base fork: basius reaches `mech_request_round`, executor POSTs signed request, receives structured 402 from a real mech HTTP server (validated; the 3 fixes were each pulled out of distinct crash modes during the run)
- [ ] No regression on on-chain path (these fixes only add to offchain executor)


packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -1613,18 +1630,28 @@ def _build_safe_tx_for_single_call(
...
+            operation=operation,
+            chain_id=self._b.params.mech_chain_id,
         )
+        if response.performative != ContractApiMessage.Performative.STATE:
+            self._logger.warning(
+                f"GnosisSafe.get_raw_safe_transaction_hash failed: "
+                f"performative={response.performative}"
+            )
+            return None
+        return cast(str, response.state.body.get("tx_hash"))
```
OjusWiZard: **Critical — deposit `tx_hash` is the raw Safe hash, not the packed settlement payload.**

This returns the bare 32-byte `get_raw_safe_transaction_hash` result. It flows unmodified into the settlement payload:
`_build_deposit_tx` → `OffchainCycleResult.tx_hash` (L810) → `MechRequestPayload.tx_hash` (request.py:1039) → `most_voted_tx_hash`.

But the on-chain path feeds the **same** `most_voted_tx_hash` field with the **packed** form — `self.tx_hex = hash_payload_to_hex(safe_tx_hash, value, SAFE_GAS, to, data, operation)` (request.py:1105 / base.py:135). The downstream `transaction_settlement` consumer decodes that packed string to recover `to/value/data/operation/gas` and execute the Safe tx. A raw 64-char hash can't be decoded into those fields, so every `DEPOSIT_NEEDED` cycle hands settlement an unexecutable tx — the deposit never settles.

The old (crashing) code called `_get_safe_tx_hash`, whose docstring claimed it "preserves the gas estimation … path the on-chain mech request already uses" — i.e. the intent was the packed format (the variable is even named `tx_hex`). The fix removed the crash but dropped the packing.

Suggested fix (mirrors `tx_hex`; `SAFE_GAS == 0` already matches your `safe_tx_gas=0`):
```python
safe_tx_hash = response.state.body.get("tx_hash")
if safe_tx_hash is None:
    return None
return hash_payload_to_hex(
    safe_tx_hash, value, SAFE_GAS, to_address, HexBytes(data).hex(), operation,
)
```

Tests missed this because the deposit tests only assert `result.tx_hash is not None` — never that it's the packed format. Add an assertion that decodes it back (or at least asserts length > 64 hex).

OjusWiZard: **Advisory — no validation on the returned hash.**

The reference (`base.py:393-401`) checks `tx_hash` is non-None, a `str`, and `len == TX_HASH_LENGTH`, logging on failure. `cast(str, ...get("tx_hash"))` instead masks a `None`/malformed body as a typed `str` and pushes the problem downstream. Worth matching base.py's validation — and this is mostly subsumed if you adopt the `hash_payload_to_hex` packing above (validate `safe_tx_hash` first).

OjusWiZard (commented):
Review of the 3 fixes. #1 (chain_id on all 11 reads) and #2 (`_pick_next_mech` gate) verified correct and complete. Fix #3 resolves the `AttributeError` but introduces a settlement-format defect — details inline.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Confirmed and fixed in 07b524c. The raw hash going into `most_voted_tx_hash` is the same bug `base.py:_build_multisend_safe_tx_hash` and `tx_hex` were already guarding against, and `skill_input_hex_to_payload` would have desersialised it to garbage. Now packed via `hash_payload_to_hex(safe_tx_hash[2:], value, SAFE_GAS, to_address, data_bytes, operation)` mirroring the on-chain path exactly. Added a round-trip decode assertion in `test_402_under_cap_returns_deposit_needed` so a future regression that drops packing fails loudly instead of leaving a raw hash on the wire.

DIvyaNautiyal07: Done in 07b524c, subsumed by the packing fix. The helper now validates `safe_tx_hash` is non-None, is a `str`, and has `len == TX_HASH_LENGTH` (matching base.py:393-401), logs a warning and returns `None` on failure. The unused `cast(str, ...)` import is removed too.

OjusWiZard (approved):
Re-reviewed after the fix commits.

- Fix #1 (chain_id on all 11 contract reads) and Fix #2 (`_pick_next_mech` gated on `use_dynamic_mech_selection`) verified correct and complete.
- Fix #3: the deposit `tx_hash` now packs via `hash_payload_to_hex` (matching the on-chain `tx_hex`) so `transaction_settlement` can decode it back to the Safe tx params, plus `None`/`isinstance`/`TX_HASH_LENGTH` validation. Packing edge cases checked (64-char strip, prefix-free `HexBytes.hex()` on pinned hexbytes 2.0.0b1).
- New test round-trips through `skill_input_hex_to_payload` — a real regression guard.
- CI all green.

Both of my prior review threads are addressed and resolved. LGTM 🚀

Minor optional follow-up (non-blocking): the token/multisend deposit test asserts only `tx_hash is not None`; a round-trip decode there would close the last coverage gap.

bennyjo (approved):
All three fixes are correct and faithfully mirror the on-chain path: `chain_id` is applied to all 11 executor contract reads (none missed), the `_pick_next_mech` gate matches `request.py:574`, and the Safe-tx helper now validates the hash and packs it via `hash_payload_to_hex(...)` like `base.py`'s `tx_hex` so the deposit payload is executable by settlement. CI green; deposit tests round-trip-decode the packed payload.

Two non-blocking nits for a follow-up: (1) bug #2's `use_dynamic_mech_selection` gate isn't covered by a discriminating test — all `_pick_next_mech` tests use empty `ranked_mechs`, so removing the gate wouldn't fail any test; a test with `use_dynamic_mech_selection=False` + non-empty `ranked_mechs` asserting the static priority is returned would lock it in. (2) The bug #3 paragraph in the description is stale — it omits the hash validation + `hash_payload_to_hex` packing that makes the deposit settleable.

Approving.

PR #99, opened 2026-07-03
refactor: consume canonical BalanceTracker + MechMarketplace, bump OA 0.21.26

DIvyaNautiyal07:
## Summary

Delete 3 slim duplicate contract packages that mech-interact was carrying and consume the canonical versions as `third_party`. Bump open-autonomy to 0.21.26 and open-aea to 2.2.9.

## Changes

**Deleted 3 slim packages:**
- `packages/valory/contracts/mech_marketplace/` — replaced by open-autonomy's extended `mech_marketplace` (adds `get_chain_id` and `get_nonce`, via valory-xyz/open-autonomy#2537)
- `packages/valory/contracts/balance_tracker_fixed_price_native/` — replaced by mech's new native-variant package (via valory-xyz/mech#464)
- `packages/valory/contracts/balance_tracker_fixed_price_token/` — replaced by mech's `balance_tracker` (now extended with `build_deposit_for_data`, via valory-xyz/mech#464)

**Rewired `offchain_request.py`** to call the canonical classes with the corresponding response key renames:

| Old call | New call | Response key rename |
|---|---|---|
| `MechMarketplace.get_balance_tracker` | `MechMarketplaceContract.get_balance_tracker_for_mech_type` | `body["balance_tracker"]` → `body["data"]` |
| `MechMarketplace.get_nonce` (slim) | `MechMarketplaceContract.get_nonce` (canonical) | `body["nonce"]` → `body["data"]` |
| `MechMarketplace.get_chain_id` (slim) | `MechMarketplaceContract.get_chain_id` (canonical) | `body["chain_id"]` → `body["data"]` |
| `BalanceTrackerFixedPriceToken.get_token` | `BalanceTrackerContract.get_token_address` | `body["token"]` → `body["token_address"]` |
| `BalanceTrackerFixedPriceToken.build_deposit_for_data` | `BalanceTrackerContract.build_deposit_for_data` (mech's) | shape unchanged |
| `BalanceTrackerFixedPriceNative.build_deposit_for_data` (no amount) | canonical `BalanceTrackerFixedPriceNativeContract.build_deposit_for_data(account, amount)` | now returns `{"data": ..., "value": amount}` to close the on-chain zero-value silent-no-op gap |

Test stubs in `test_offchain_request_helpers.py` updated to match the new response keys (29 replacements).

**Version bumps:**
- `open-autonomy` 0.21.23 → 0.21.26
- `open-aea*` 2.2.7 → 2.2.9 across `pyproject.toml` and 21 package YAMLs that pin the ledger plugin

## Deferred locking (draft)

- **`uv lock`** — currently fails resolving `open-autonomy==0.21.26` because it isn't on PyPI yet. open-autonomy #2537 is approved and merging shortly; will regen this lock immediately after the release cuts.
- **`autonomy packages sync`** — currently times out fetching mech's new `balance_tracker` hash (`bafybeie7xvhpyztpixha7kjx6z6deyvvsk5er4xpdkpgqc65q552xpvi6e`). Waiting for the post-merge IPFS push to settle, then will run `autonomy packages lock` to regenerate `packages.json` and `skill.yaml` fingerprints.

Both are mechanical follow-ups; every hash referenced in this diff is what will be on IPFS once those two upstreams settle. Marking as draft until both regens are clean.

## Test plan

- [ ] `uv lock --check` green after OA 0.21.26 tags
- [ ] `autonomy packages sync` completes without timeout
- [ ] `autonomy packages lock --check` green
- [ ] Existing offchain e2e test (Anvil Base fork + Pearl-deployed Gnosis mech at service 2692) passes

## Related

- valory-xyz/open-autonomy#2537 — adds `get_chain_id` / `get_nonce` to canonical `mech_marketplace`
- valory-xyz/mech#464 — extends `balance_tracker` with `build_deposit_for_data`, adds new `balance_tracker_fixed_price_native` package (merged into `offchain-epic`)


bennyjo (commented):
Two blocking contract-signature mismatches against the canonical classes, plus a test-stub gap that hides them.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -1109,7 +1111,7 @@ def _read_on_chain_nonce(self) -> Generator[None, None, Optional[int]]:
                 f"mapNonces read failed: performative={response.performative}"
             )
             return None
-        nonce = response.state.body.get("nonce")
+        nonce = response.state.body.get("data")
```
bennyjo: Line 1106 passes `sender=` but the canonical `get_nonce` expects `sender_address=`, so this call errors, `_read_on_chain_nonce` returns `None`, and every offchain request short-circuits to `OFFCHAIN_ALL_FAILED` at line 700-705.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -63,15 +63,17 @@
...
 
-from packages.valory.contracts.balance_tracker_fixed_price_native.contract import (
-    BalanceTrackerFixedPriceNative,
+from packages.valory.contracts.balance_tracker.contract import (
+    BalanceTrackerContract as BalanceTracker,
 )
-from packages.valory.contracts.balance_tracker_fixed_price_token.contract import (
-    BalanceTrackerFixedPriceToken,
+from packages.valory.contracts.balance_tracker_fixed_price_native.contract import (
+    BalanceTrackerFixedPriceNativeContract as BalanceTrackerFixedPriceNative,
```
bennyjo, line 70: The canonical native `build_deposit_for_data` requires an `amount` arg, but the call site at lines 1497-1504 still passes only `account=` — so it errors and the native deposit path breaks, contrary to the PR description's claim that it "now passes amount."

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -829,8 +829,8 @@ class TestFreshCycle:
     def _native_reads(self) -> List[Any]:
         """Sequence of contract reads a single happy-path attempt consumes."""
         return [
-            _state_resp({"chain_id": 100}),  # _resolve_chain_id_int
-            _state_resp({"nonce": 7}),  # _read_on_chain_nonce
+            _state_resp({"data": 100}),  # _resolve_chain_id_int
```
bennyjo, line 918: The stub's `get_contract_api_response(**kwargs)` (line 745) discards all kwargs, so nothing checks the call arguments against the canonical signatures — which is why both mismatches above pass CI.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
OjusWiZard, line 70: **[Critical] Native auto-deposit breaks: `_build_native_deposit_tx` (L1497-1504, no hunk in this diff) never passes the now-required `amount`.**

The canonical `balance_tracker_fixed_price_native` at the hash this PR pins (mech commit `2aff76fb`, "address PR review — native amount guard") declares `amount: int` as a **required** parameter on `build_deposit_for_data` and raises `ValueError` for `amount <= 0`:

```python
def build_deposit_for_data(cls, ledger_api, contract_address, account: str, amount: int) -> JSONLike:
    if amount <= 0:
        raise ValueError(...)
    ...
    return {"data": ..., "value": amount}
```

But the call site at L1497-1504 still passes only `account=self._safe_address()` — no `amount` kwarg. Once `autonomy packages sync` fetches the pinned version, every native-path 402 deposit fails with a missing-argument error → non-STATE performative → `return None` → the whole native deposit-and-retry flow dies. The token sibling `_build_token_deposit_multisend` (L1556) passes `amount=deposit_amount` correctly.

The PR description's own table documents the new `(account, amount)` signature — the description was updated but this call site wasn't. Careful when verifying: a mech checkout at `34bab1d7` (one commit earlier, "drop amount param") shows an amount-less signature; the *pinned* final state re-added it.

Fix: add `amount=deposit_amount` at L1502. Keeping `value=deposit_amount` on the Safe envelope (L1523) stays correct; consuming the returned `"value"` key is optional.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
OjusWiZard: **[Important] Both failure branches here are silent.** `nonce is None` and the `int()` `TypeError`/`ValueError` branch return `None` without logging, while the sibling `_resolve_chain_id_int` (L1144-1150) — reading the same generic `"data"` key — warns on exactly the same failure shapes. A degraded RPC response or wrapper regression collapses into the generic `OFFCHAIN_TIMEOUT_ALL_MECHS` with no diagnostic trail: an operator can't tell a performative mismatch (logged) from a missing `data` key or non-numeric nonce (both unlogged). Mirror the warning pattern used in `_resolve_chain_id_int` / `_resolve_delivery_rate` / `_resolve_canonical_tracker`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -1247,7 +1249,7 @@ def _resolve_canonical_tracker(
                 f"performative={response.performative}"
             )
             return None
-        raw = response.state.body.get("balance_tracker")
+        raw = response.state.body.get("data")
```
OjusWiZard: **[Important, structural] This canonicalization made 3 helpers converge on the identical shape** — `_read_on_chain_nonce`, `_resolve_chain_id_int`, and `_resolve_canonical_tracker` now all do GET_STATE → performative check → `body.get("data")` → coerce, and 8 more sites in this file share the pattern with other keys (11 near-identical guard blocks total). A shared `_read_contract_data(contract_address, contract_id, contract_callable, error_label, **kwargs)` helper — mirroring the `PaymentChallenge`/`_try_int` adapter this file already has at the HTTP 402 boundary — would leave only per-site coercion at each call site, and make the *next* contract-wrapper key rename fail once at the adapter instead of silently at N call sites. That's precisely the bug class the native-deposit Critical in this PR fell into: the only thing that caught the 7 renames here was updating tests in lockstep.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -1276,21 +1278,20 @@ def _resolve_tracker_token(
...
                 f"{tracker_address}: performative={response.performative}"
             )
             return None
-        raw = response.state.body.get("token")
+        raw = response.state.body.get("token_address")
         if not isinstance(raw, str) or not raw:
             self._logger.warning(
-                "BalanceTrackerFixedPriceToken.token() returned an "
-                f"unexpected value {raw!r}"
+                "BalanceTracker.token() returned an " f"unexpected value {raw!r}"
```
OjusWiZard: **[Advisory] Rename artifact:** two adjacent string literals where only the second is an f-string — `"BalanceTracker.token() returned an " f"unexpected value {raw!r}"`. Functionally identical via implicit concatenation, but reads oddly; fold into a single f-string.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -268,9 +268,21 @@ def _parse(self, response: Any) -> _PollSnapshot:
...
+                    # (target.error becomes "Unknown"); flag it so the
+                    # drift is visible rather than a fake failed prediction.
+                    self._logger.warning(
+                        "Offchain 'ok' envelope missing 'result' key; "
+                        "keys=%s. Treating as a failed poll.",
+                        sorted(envelope.keys()),
+                    )
+                inner_result = envelope.get("result")
+            else:
+                inner_result = envelope
```
OjusWiZard, line 297: **[Important] The schema-drift warning above doesn't cover the likeliest truncation case.** If the payload is `{"status": "ok"}` with no `"response"` key at all (or `"response"` is a non-dict scalar), `envelope` is `None`/non-dict, `isinstance` fails, and this branch silently sets `inner_result = envelope` — flowing into `_apply_snapshot` as `result=None`, `error="Unknown"`, indistinguishable from a routine tool failure with zero log trace. That's the same failure mode the `missing 'result' key` warning was added to expose (f4dccfa targeted "renamed field, truncated payload, new nesting"), but only the dict-missing-`result` shape gets the warning. Suggest also warning here when `envelope` is `None` or not a dict, e.g. `Offchain 'ok' status but 'response' envelope missing/malformed: %r`.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -879,7 +879,7 @@ def test_402_under_cap_returns_deposit_needed(self) -> None:
             contract_api_responses=[
                 *self._native_reads(),
                 # _validate_402_destination → marketplace.get_balance_tracker
```
OjusWiZard: **[Important] Stale comment:** `marketplace.get_balance_tracker` no longer exists — the callable this fixture mocks is now `get_balance_tracker_for_mech_type` (the payload on the next line was updated to `{"data": ...}`, the comment wasn't). Same at L1487. Grepping for `get_balance_tracker` to trace this fixture back to production code now finds nothing.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -879,7 +879,7 @@ def test_402_under_cap_returns_deposit_needed(self) -> None:
...
-                _state_resp({"balance_tracker": canonical_tracker}),
+                _state_resp({"data": canonical_tracker}),
                 # _build_native_deposit_tx → BalanceTracker.build_deposit_for_data
                 _state_resp({"data": b"\x01\x02\x03"}),
```
OjusWiZard, line 970: **[Important] Two test gaps that let the native-deposit Critical through:**

1. `_StubBehaviour.get_contract_api_response` ignores all kwargs and pops canned responses off a list, so the missing `amount` kwarg on the real `build_deposit_for_data` call is invisible to every native-deposit test. Add a kwargs assertion on `build_deposit_for_data` calls (native *and* token paths), analogous to the existing `_approve_amount` helper that inspects `build_approval_tx` calls.
2. The deposit `value` is never asserted: these tests decode the packed Safe tx via `skill_input_hex_to_payload` but check only `safe_tx_hash`/`to_address`, never `decoded["ether_value"]`. A regression hard-coding `value=0` — the exact zero-value silent-no-op class this PR closes upstream — would pass every existing test. Add `decoded["ether_value"] == deposit_amount`.

(Related: the stub's `_get_safe_tx_hash`/`safe_tx_calls` recording machinery is dead code — production calls `get_contract_api_response(contract_callable="get_raw_safe_transaction_hash", ...)` directly.)

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -1481,14 +1481,14 @@ def test_pay_to_mismatch_treated_as_bad_response(self) -> None:
                 _FakeMechInfo(mech_b, "https://mech-bb.example"),
             ],
             contract_api_responses=[
-                _state_resp({"chain_id": 100}),
-                _state_resp({"nonce": 7}),
+                _state_resp({"data": 100}),
+                _state_resp({"data": 7}),
                 *per_attempt_reads,  # attempt #1
                 # validation → marketplace.get_balance_tracker
```
OjusWiZard: **[Important] Stale comment** (same as L881): the callable is now `get_balance_tracker_for_mech_type`, not `marketplace.get_balance_tracker`.

OjusWiZard (commented):
Multi-perspective review (code, comments, tests, silent failures, type design, simplification). One **Critical** — anchored inline on the native import since `_build_native_deposit_tx` itself has no diff hunk — plus Important test/logging gaps, all inline. Two findings that have no diff anchor:

**[Important] `.gitignore` missing entries for the three newly third-party packages.** Lines 16-22 ignore the seven pre-existing third-party contracts (`agent_registry` … `service_registry`), but `mech_marketplace`, `balance_tracker`, and `balance_tracker_fixed_price_native` — moved to `third_party` in this PR's `packages.json` — have no entries. After `autonomy packages sync --update-packages`, all three show up untracked and risk being committed, contrary to the repo convention that third-party packages are synced, not vendored.

**[Advisory] Collapsing `"nonce"`/`"chain_id"`/`"balance_tracker"` into a uniform `"data"` key** removed an incidental discriminator: previously a misrouted fixture/response returned `None` for the wrong key and got caught; now the two int reads are indistinguishable to `isinstance`/`int()` guards, and the ordered `_state_resp({"data": ...})` fixture lists have no field name left to catch a swap. Not an active bug (dialogue matching pairs responses, not order), but worth a `_state_resp({})` (STATE performative, missing `data`) test to pin the defensive branches — currently only the ERROR-performative path is exercised.

Verification notes: the response-key renames were checked against the canonical contract sources at the pinned mech commit (`2aff76fb`) and are all correct; both changed test files pass (86 passed) after `autonomy packages sync`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 0b176a6. Renamed the kwarg from `sender=` to `sender_address=` to match the canonical get_nonce signature. Also caught locally via aea-helpers run-agent (the whole offchain path was short-circuiting to OFFCHAIN_ALL_FAILED because _read_on_chain_nonce silently returned None on the kwarg-mismatch).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 0b176a6 — build_deposit_for_data now passes amount=deposit_amount to match the canonical signature (`account: str, amount: int`). PR description was accurate for the token path but stale on native; native was still passing only account=.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Right — the stub was intentionally permissive when it was mostly asserting sequences of canned responses, but the two mismatches above proved that permissiveness was hiding real bugs. Tightened in 0b176a6 via a per-callable kwargs allowlist: `_CANONICAL_REQUIRED_KWARGS` maps each contract_callable to the kwargs the canonical wrapper actually requires, and the stub now raises AssertionError if any are missing. Same shape as the existing _approve_amount helper. The get_nonce sender/sender_address mismatch and the native-deposit missing amount both now fail these tests deterministically.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 0b176a6. Confirmed against the pinned mech contract at v0.34.2-rc2 — build_deposit_for_data raises ValueError on amount <= 0 and the encoded calldata plus value are both propagated. Call site (was L1497-1504) now passes `amount=deposit_amount` and reads both data and value from the response body.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 1fd6ecd. All 11 GET_STATE call sites in offchain_request.py now go through a shared _read_contract_state helper that owns the performative check + missing-key branch with a consistent warning shape. Each caller keeps ownership of coercion / shape validation on the returned raw. The three canonicalised helpers (_read_on_chain_nonce, _resolve_chain_id_int, _resolve_canonical_tracker) now all log at the same shape, and a degraded RPC response or wrapper regression surfaces at the adapter.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Done in 1fd6ecd. The 11 near-identical guard blocks all now call `_read_contract_state(contract_address, contract_id, contract_callable, error_label, **kwargs)`. The next canonical-wrapper key rename will fail at one place (the adapter) rather than silently at N call sites — exactly the failure mode of the native-deposit `amount` regression this PR fell into.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 1fd6ecd — folded the two adjacent literals in _resolve_tracker_token into a single f-string. Grep for adjacent literal + f-string in the file returns clean.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in 1fd6ecd. Added a warning on the else-branch as well (envelope is None / non-dict scalar): 'Offchain ok status but response envelope missing/malformed: %r. Treating as a failed poll.' Now the {"status": "ok"} truncation case leaves a diagnostic trail equivalent to the missing-result-key branch instead of silently producing error="Unknown".

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Fixed in 1fd6ecd. Both stale comments updated: 'marketplace.get_balance_tracker' -> 'marketplace.get_balance_tracker_for_mech_type' at the L881 fixture and the L1487 test (which is now at the corrected lines 942 and 1560 respectively). Grep for the old name returns zero.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Both gaps closed in 0b176a6:
1. _StubBehaviour.get_contract_api_response now asserts required kwargs per contract_callable via _CANONICAL_REQUIRED_KWARGS (allowlist per callable). build_deposit_for_data is in there for both native (`account, amount`) and token (`account, amount`) — the missing amount now fails the test deterministically.
2. Added assertions on the returned value key on the native path (value == deposit_amount) so a caller spreading the dict cannot omit it and hit the on-chain silent-no-op path.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Same fix as 3520760230 — updated in 1fd6ecd.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -1088,6 +1103,45 @@ def _sign_request_id(
...
+            )
+            return None
+        body = response.state.body or {}
+        if result_key not in body:
+            self._logger.warning(
+                f"{error_label} response missing {result_key!r} key; "
+                f"body keys={sorted(body.keys())}"
+            )
+            return None
+        return body[result_key]
```
OjusWiZard: **[Important] Present-but-`None` values are now *quieter* than before the refactor.** The helper warns when `result_key` is absent, but `{"data": None}` returns `None` here with no warning — and the migration's caller-side `if raw is None: return None` short-circuits now run *before* the `isinstance`/`except` logging that used to catch exactly this shape at 8-10 pre-refactor sites (chain-id, payment-type, delivery-rate, tracker, token, both deposit builders, multisend `get_tx_data`, and `_build_safe_tx_for_single_call`, which sits on every settlement path — its old combined `is None or not isinstance(...)` check always logged). Behavior is still safe, but a degraded RPC returning `{"data": null}` becomes undiagnosable — the exact property this docstring says the extraction fixed.

One-line fix that repairs all call sites at once: also warn (and return `None`) when `body[result_key] is None`. None of the `TestReadContractState` tests cover a `None`-valued key, so add one alongside.

(Also, a docstring nit while here: "silently at each call site" overstates the before-state — only the `mapNonces` read was truly silent; the other ten warned via coercion failures, just inconsistently.)

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -1088,6 +1103,45 @@ def _sign_request_id(
...
+        error_label: str,
+        result_key: str = "data",
+        **kwargs: Any,
+    ) -> Generator[None, None, Optional[Any]]:
+        """Shared GET_STATE adapter for the 11 near-identical read blocks.
+
+        Handles the performative check and the ``body[result_key]`` lookup
+        with a consistent warning pattern so a degraded RPC or a
+        contract-wrapper key rename fails loudly at one place instead of
+        silently at each call site (review 3520760220). Callers keep
```
OjusWiZard: **[Advisory] Bare numeric review IDs are unresolvable for future readers.** `(review 3520760220)` here — and `Regression 3520760216` in `test_offchain_request_helpers.py` (`TestReadContractState`) — are GitHub review/comment IDs with no URL, and this is the only place in the repo using that convention; everything else uses the `C<N>` short-codes or a named form (`review: bennyjo on PR #95`). Either switch to the established convention or drop the parenthetical — the docstring's rationale stands on its own.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -547,6 +558,7 @@ def from_dict(cls, raw: Dict[str, Any]) -> Optional["PendingRequest"]:
                 delivery_rate=int(raw["delivery_rate"]),
                 ipfs_hash=str(raw["ipfs_hash"]),
                 ipfs_data=str(raw["ipfs_data"]),
+                metadata_nonce=str(raw.get("metadata_nonce", "")),
```
OjusWiZard: **[Important] `metadata_nonce: null` coerces to the literal string `"None"`.** The two-arg `raw.get("metadata_nonce", "")` only substitutes the default when the key is *absent* — a present-but-`null` value (e.g. an upstream producer emitting `nonce: null` into `MechMetadata`, which has no runtime type enforcement, threaded through `_build_pending` → `to_json`) comes back as `None`, and `str(None)` produces `"None"`. `__post_init__` validates `request_id`/`mech_address`/`nonce`/`delivery_rate` but not `metadata_nonce`, so the corrupted value passes silently and ships downstream as the correlation key in `_serialise_pending_response`.

Fix: `metadata_nonce=str(raw.get("metadata_nonce") or "")`, plus a `metadata_nonce` check in `__post_init__` consistent with the other fields.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -742,9 +774,38 @@ def __init__(
...
         self.safe_tx_calls: List[Dict[str, Any]] = []
 
+    # Required kwargs per canonical contract callable, on top of the framework-
+    # supplied set (performative, contract_address, contract_id, chain_id).
+    # Every callable exercised by ``OffchainRequestExecutor`` is listed here;
+    # if a call site drops one of these, the stub raises instead of silently
+    # returning the queued response -- which is how bennyjo's ``sender=``
+    # vs ``sender_address=`` and missing-``amount`` regressions passed CI
+    # before this stub was tightened.
+    _CANONICAL_REQUIRED_KWARGS = {
```
OjusWiZard, line 798: **[Important] `get_raw_safe_transaction_hash` is missing from this map** — and since the stub's guard is `if expected is not None`, the unlisted callable silently skips validation entirely. That exempts the single highest-value call site (`_build_safe_tx_for_single_call`, the Safe settlement tx hash) from the exact kwarg-rename/kwarg-drop regression class this map was built to catch, and contradicts the comment's claim that every callable exercised by `OffchainRequestExecutor` is listed. Two independent review passes converged on this one.

Fix: add `"get_raw_safe_transaction_hash": {"to_address", "value", "data", "safe_tx_gas", "operation"}`, plus a `TestStubValidatesCanonicalKwargs` case mirroring `test_missing_sender_address_on_get_nonce_raises`. Longer-term caveat worth a comment on the map: it's a hand-maintained shadow of the real wrapper signatures — if upstream adds a required parameter, this dict won't know.

OjusWiZard (commented):
Follow-up review of the fix commits (`0b176a6`, `1fd6ecd`, `724248a`). The big items verified clean: the v0.34.2-rc2 hash re-pin is byte-identical to the upstream tag (`autonomy packages lock --check` passes, every callable signature and response key matches, including native `build_deposit_for_data(account, amount)`); the metadata-nonce wiring is correct on both fresh-cycle and retry paths; the `_read_contract_state` migration has no control-flow or kwargs drift; 97/97 tests pass. Remaining findings inline — no Criticals.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e2dfb0a. `_read_contract_state` now emits "$error_label response 'data' value is None" when the key is present with a null value, restoring parity with the pre-refactor sites that inherited this shape via their combined `is None or not isinstance(...)` guards. Added `test_present_but_null_value_returns_none_with_warning` in TestReadContractState to pin it (asserts both the value-is-None warning and the error_label are in the log).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e2dfb0a. Dropped both bare numeric IDs — `(review 3520760220)` from the `_read_contract_state` docstring rationale (the rationale stands on its own) and `Regression 3520760216:` from the test name (renamed to `test_non_state_performative_returns_none_with_warning` with the docstring describing the property directly). Grep for either ID returns clean.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in e2dfb0a. Two-part fix: (1) coercion in `from_dict` now uses `str(raw.get("metadata_nonce") or "")` so a present-but-null value falls into the same empty-string bucket as a missing key; (2) added an `isinstance(self.metadata_nonce, str)` guard in `__post_init__` mirroring the other field validations. Added `test_null_metadata_nonce_normalises_to_empty` to pin the coercion and confirm the "None" literal can't ship as a correlation key.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Fixed in e2dfb0a. Added `"get_raw_safe_transaction_hash": {"to_address", "value", "data", "safe_tx_gas", "operation"}` to `_CANONICAL_REQUIRED_KWARGS`, plus `test_missing_kwargs_on_get_raw_safe_transaction_hash_raises` that parametrises over all 5 kwargs and asserts the stub raises when each is dropped in turn — mirroring the existing sender_address test but covering the whole kwarg-drop surface for the Safe-tx-hash build, since that call site sits on every settlement path. Comment on the map above updated to match.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -760,14 +783,15 @@ def _fresh_cycle(self) -> Generator[None, None, OffchainCycleResult]:
                 delivery_rate=delivery_rate,
                 ipfs_hash=ipfs_hash,
                 ipfs_data=ipfs_data,
+                metadata_nonce=request_meta.nonce,
```
OjusWiZard: **[Important] Reproduced crash: the new `metadata_nonce` guard can't fire where it was aimed, and crashes here instead.**

Via `from_dict` the `__post_init__` isinstance check (L534-538) is provably unreachable — `str(raw.get("metadata_nonce") or "")` always hands it a string. Its only live path is this direct construction, which passes `request_meta.nonce` with no coercion and no try/except anywhere up through `_run_offchain_request_cycle`.

`MechMetadata` is a plain dataclass built by the producer skill via `MechMetadata(**metadata_item)` with no runtime enforcement, and this module already treats `nonce=None` as legitimate — `build_request_metadata(nonce_str=...)`, called just above, explicitly falls back to a generated UUID for `None`. Reproduced: driving `_fresh_cycle()` with `MechMetadata(nonce=None)` raises an uncaught `ValueError` — and it fires *after* `_post_signed_request` has already POSTed the signed request to the mech, i.e., after an irreversible side effect. Pre-guard, the same input was a quiet shape bug (`"None"` string); now it's a mid-cycle crash instead of a graceful `OFFCHAIN_ALL_FAILED`.

Fix: coerce here to match `from_dict`'s tolerance — `metadata_nonce=str(request_meta.nonce or "")` — and add a test driving `_fresh_cycle` with `MechMetadata(nonce=None)`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -547,6 +563,13 @@ def from_dict(cls, raw: Dict[str, Any]) -> Optional["PendingRequest"]:
...
                 ipfs_data=str(raw["ipfs_data"]),
+                # `str(None)` produces the literal "None" — a present-but-null
+                # key (e.g. an upstream producer emitting `metadata_nonce:
+                # null`, which MechMetadata doesn't runtime-enforce) would
+                # otherwise ship as the correlation key in
+                # `_serialise_pending_response`. The `or ""` normalises None
+                # / missing to the same empty-string fallback.
+                metadata_nonce=str(raw.get("metadata_nonce") or ""),
             )
         except (KeyError, TypeError, ValueError):
```
OjusWiZard, line 574: **[Important] This `except` silently drops a corrupt stored pending request — and the strengthened `__post_init__` validation widens that funnel.**

No logging here, and `run()` (L681-691) can't distinguish "nothing was pending" (normal) from "a pending request was stored but failed validation" (corruption) — both fall through to `_fresh_cycle()`. Since the pending request is the bridge between "deposit just paid for this request_id" and "resume the POST", a corrupt dict here means the executor silently abandons the paid-for correlation and starts over, with zero diagnostic trail.

Same pattern in `offchain_response.py` `_load_pending` (L347-358), where the existing warning text ("no pending request on synced data") is actively misleading for the corruption case.

Fix: log in this branch (exception type + message; keys, not full values), or have `_load_pending_request` distinguish empty-raw from failed-validation so `run()` can warn accurately.

OjusWiZard (approved):
Approving — third review round. The delta since last round verified clean on all the load-bearing checks: `autonomy packages lock --check` passes against the relocked fingerprints, the `get_raw_safe_transaction_hash` kwargs map entry matches the production call site exactly, the mypy fix is genuine narrowing (no cast/ignore laundering), the present-but-null warning is implemented and tested, and both suites pass 100/100.

Two pre-merge asks inline (a reproduced crash path and its sibling logging gap), plus one item that has no diff line to anchor to:

**[Important] `.gitignore` is still missing entries for the three newly third-party packages** — `packages/valory/contracts/mech_marketplace`, `balance_tracker`, and `balance_tracker_fixed_price_native`. The seven pre-existing third-party contracts all have entries (`.gitignore:16-22`); after `autonomy packages sync --update-packages` these three sit untracked in `git status` (confirmed live), one `git add .` away from vendoring synced artifacts, contrary to the repo convention that third-party packages are synced, not vendored. Three lines to add.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Verified and fixed in 62e8388. Reproduced against the actual code path — you were right on every point:

- `MechMetadata.nonce` is typed `str` but the dataclass has no runtime enforcement; `states/base.py:455` (`MechMetadata(**metadata_item)` from the DB blob) reaches the direct-construction path with `nonce=None` if the producer emits `{"nonce": null}`.
- Timing confirmed: `_post_signed_request` at L765 runs *before* `_build_pending` at L778. The isinstance guard would raise mid-cycle past the irreversible POST.
- `from_dict`'s guard IS provably unreachable — `str(... or "")` always hands a string.

Fix as suggested: `metadata_nonce=str(request_meta.nonce or "")` at L786, matching `build_request_metadata`'s upstream fallback. Added `test_null_metadata_nonce_does_not_crash_after_signed_post` in TestFreshCycle that drives `_fresh_cycle` with `MechMetadata(**{"prompt": "x", "tool": "t", "nonce": None})` — the `**` unpack bypasses the type checker the way the production DB path does. Asserts DONE completion + `metadata_nonce == ""` (not the literal "None").

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Verified and fixed in 62e8388. Both loaders now log a validation warning when raw is truthy but `from_dict` returns None — logging keys (not values) so a corrupt payload doesn't spill request_id / signature material.

- `_load_pending_request` (offchain_request.py:1756): warning emitted, run() proceeds to `_fresh_cycle` as before but now with a diagnostic trail.
- `_load_pending` (offchain_response.py:347): same shape. The outer "no pending request on synced data" warning at L114 is no longer misleading — the inner corruption warning fires first in the failure case, and the outer only fires on the genuine empty-raw path.

Added `TestLoadPendingRequestDistinguishesCorruption` with two cases: (1) `corrupt = {"mech_url": ..., "sender": ...}` (missing `request_id`, KeyError inside `from_dict`) → warning fires; (2) empty raw → no warning.

PR #101, opened 2026-07-06
fix(offchain): revert RESPONSE_ROUND_TIMEOUT to shared ROUND_TIMEOUT

DIvyaNautiyal07:
## Summary

- Reverts the dedicated `RESPONSE_ROUND_TIMEOUT` event: the response round times out via the shared `Event.ROUND_TIMEOUT` again, so the flag-off FSM is byte-for-byte identical to main and each consumer keeps governing the timeout with its existing override (trader 1800s, market-resolver 1200s, optimus/meme-ooorr 30s framework default).
- Why revert: the `SharedState.setup` rebind mutated the class-level `event_to_timeout` after consumers' `chain()` had already copied timeouts, so the dedicated event never took effect in composed deployments. Removed the rebind entirely.
- The flag-ON safety net moves to runtime: `_run_offchain_response_cycle` now reads the effective timeout off the live composed app (`context.state.round_sequence.abci_app.event_to_timeout`) and logs an error when it cannot fit the off-chain poll budget (`offchain_poll_timeout_seconds` + 30s overhead, 330s with defaults). Reading the live app is what makes consumer rebinds visible.
- Bumps the open-autonomy pin in `skill.yaml` to `==0.21.26` to match the framework version.

## Release notes for the epic

- Consumers repin mech-interact and drop any `RESPONSE_ROUND_TIMEOUT` references in the same release (atomic upgrade). The only known reference outside this repo is a stale comment block in meme-ooorr `memeooorr_chained_abci/composition.py`.
- Deployments enabling `use_offchain` need an effective mech-interact round timeout of at least `offchain_poll_timeout_seconds + 30s` (330s with defaults); the new runtime guard logs an error otherwise.
- After merge to main, cut the final non-RC release.

## Test plan

- [x] New unit tests: response round maps `ROUND_TIMEOUT` to `FinishedMechResponseTimeoutRound`, `event_to_timeout` only carries the shared event, `RESPONSE_ROUND_TIMEOUT` no longer exists on the enum
- [x] New guard tests: error logged below budget, silent at/above budget and when the timeout key is absent
- [x] Full skill suite passes (524 tests); all CI workflow checks run locally (hashes, specs, docstrings, lint, type checks, security scans)

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/base.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/base.py
@@ -49,7 +49,10 @@ from packages.valory.skills.mech_interact_abci.models import (
     MultisendBatch,
     SharedState,
 )
-from packages.valory.skills.mech_interact_abci.states.base import SynchronizedData
+from packages.valory.skills.mech_interact_abci.states.base import (
+    Event,
+    SynchronizedData,
+)
 from packages.valory.skills.transaction_settlement_abci.payload_tools import (
     hash_payload_to_hex,
 )
@@ -62,6 +65,11 @@ WaitableConditionType = Generator[None, None, bool]
 # which is what we want in most cases
 # more info here: https://safe-docs.dev.gnosisdev.com/safe/docs/contracts_tx_execution/
 SAFE_GAS = 0
+
+# Slack on top of the off-chain poll budget for payload assembly and
+# consensus; used to compute the minimum round timeout enforced by
+# ``_check_round_timeout_fits_poll_budget``.
+OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS = 30.0
 
 
 class MechInteractBaseBehaviour(BaseBehaviour, ABC):
@@ -309,6 +317,37 @@ class MechInteractBaseBehaviour(BaseBehaviour, ABC):
             msg = f"Retrying in {self.params.mech_interaction_sleep_time} seconds."
             self.context.logger.info(msg)
             yield from self.sleep(self.params.mech_interaction_sleep_time)
+
+    def _check_round_timeout_fits_poll_budget(self) -> None:
+        """Fail fast when the round timeout cannot fit the off-chain poll budget.
+
+        The off-chain poll runs inside ``MechResponseRound``, bounded by the
+        composed app's ``ROUND_TIMEOUT`` for this skill's rounds. That value
+        is owned by the consumer repo's round-timeout override, so it cannot
+        be validated from params alone; read the effective value off the live
+        app instead. A timeout below ``offchain_poll_timeout_seconds`` plus
+        overhead means every off-chain request times out mid-poll, so such a
+        deployment is never legitimate and refusing to run beats degrading.
+
+        :raises ValueError: when ``use_offchain`` is enabled and the effective
+            round timeout is below the poll budget.
+        """
+        poll_budget = (
+            self.mech_marketplace_config.offchain_poll_timeout_seconds
+            + OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS
+        )
+        effective = self.shared_state.round_sequence.abci_app.event_to_timeout.get(
+            Event.ROUND_TIMEOUT
+        )
+        if effective is not None and effective < poll_budget:
+            raise ValueError(
+                f"use_offchain is enabled but the effective mech-interact round timeout "
+                f"({effective}s) is below the off-chain poll budget "
+                f"(offchain_poll_timeout_seconds + {OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS}s "
+                f"overhead = {poll_budget}s). The response round would time out mid-poll "
+                f"on every request. Raise the consumer's round-timeout override to at "
+                f"least {poll_budget}s or lower offchain_poll_timeout_seconds."
+            )
 
     def finish_behaviour(self, payload: BaseTxPayload) -> Generator:  # pragma: no cover
         """Finish the behaviour."""
--- a/packages/valory/skills/mech_interact_abci/behaviours/base.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/base.py
@@ -321,13 +321,14 @@ class MechInteractBaseBehaviour(BaseBehaviour, ABC):
     def _check_round_timeout_fits_poll_budget(self) -> None:
         """Fail fast when the round timeout cannot fit the off-chain poll budget.
 
-        The off-chain poll runs inside ``MechResponseRound``, bounded by the
-        composed app's ``ROUND_TIMEOUT`` for this skill's rounds. That value
-        is owned by the consumer repo's round-timeout override, so it cannot
-        be validated from params alone; read the effective value off the live
-        app instead. A timeout below ``offchain_poll_timeout_seconds`` plus
-        overhead means every off-chain request times out mid-poll, so such a
-        deployment is never legitimate and refusing to run beats degrading.
+        The off-chain HTTP cycles run inside this skill's rounds
+        (``MechRequestRound`` and ``MechResponseRound``), bounded by the
+        composed app's ``ROUND_TIMEOUT``. That value is owned by the consumer
+        repo's round-timeout override, so it cannot be validated from params
+        alone; read the effective value off the live app instead. A timeout
+        below ``offchain_poll_timeout_seconds`` plus overhead means every
+        off-chain cycle times out mid-poll, so such a deployment is never
+        legitimate and refusing to run beats degrading.
 
         :raises ValueError: when ``use_offchain`` is enabled and the effective
             round timeout is below the poll budget.
@@ -344,8 +345,9 @@ class MechInteractBaseBehaviour(BaseBehaviour, ABC):
                 f"use_offchain is enabled but the effective mech-interact round timeout "
                 f"({effective}s) is below the off-chain poll budget "
                 f"(offchain_poll_timeout_seconds + {OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS}s "
-                f"overhead = {poll_budget}s). The response round would time out mid-poll "
-                f"on every request. Raise the consumer's round-timeout override to at "
+                f"overhead = {poll_budget}s). The round running the off-chain cycle "
+                f"would time out mid-poll on every request. "
+                f"Raise the consumer's round-timeout override to at "
                 f"least {poll_budget}s or lower offchain_poll_timeout_seconds."
             )
 
--- a/packages/valory/skills/mech_interact_abci/behaviours/response.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/response.py
@@ -38,6 +38,7 @@ from packages.valory.skills.mech_interact_abci.behaviours.request import V1_HEX_
 from packages.valory.skills.mech_interact_abci.models import MechResponseSpecs, Ox
 from packages.valory.skills.mech_interact_abci.payloads import JSONPayload
 from packages.valory.skills.mech_interact_abci.states.base import (
+    Event,
     MECH_RESPONSE,
     MechInteractionResponse,
     MechRequest,
@@ -53,6 +54,9 @@ BYTES32_HEX_FORMAT_SPEC = "064x"
 HEX_BASE = 16
 ADDRESS_ZERO = "0x0000000000000000000000000000000000000000"
 DELIVERY_MECH_INDEX = 1
+# Slack on top of the off-chain poll budget for payload assembly and
+# consensus, so the round timeout never fires mid-poll.
+OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS = 30.0
 
 
 class MechResponseBehaviour(MechInteractBaseBehaviour):
@@ -611,6 +615,36 @@ class MechResponseBehaviour(MechInteractBaseBehaviour):
                     f"There was an error in the mech's response: {self.current_mech_response.error}"
                 )
 
+    def _warn_if_round_timeout_below_poll_budget(self) -> None:
+        """Warn loudly when the response-round timeout cannot fit the poll budget.
+
+        The off-chain poll runs inside ``MechResponseRound``, bounded by the
+        composed app's ``ROUND_TIMEOUT`` for this skill's rounds. That value
+        is owned by the consumer repo (e.g. trader's
+        ``mech_interact_round_timeout_seconds`` rebind), so it cannot be
+        validated from params alone; read the effective value off the live
+        app instead. If it is below ``offchain_poll_timeout_seconds`` plus
+        overhead, the round times out mid-poll and every off-chain request
+        slower than the round timeout lands in
+        ``FinishedMechResponseTimeoutRound``.
+        """
+        poll_budget = (
+            float(self.mech_marketplace_config.offchain_poll_timeout_seconds)
+            + OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS
+        )
+        effective = self.context.state.round_sequence.abci_app.event_to_timeout.get(
+            Event.ROUND_TIMEOUT
+        )
+        if effective is not None and effective < poll_budget:
+            self.context.logger.error(
+                f"use_offchain is enabled but the effective mech-interact round timeout "
+                f"({effective}s) is below the off-chain poll budget "
+                f"(offchain_poll_timeout_seconds + {OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS}s "
+                f"overhead = {poll_budget}s). The response round will time out mid-poll. "
+                f"Raise the consumer's round-timeout override to at least {poll_budget}s "
+                f"or lower offchain_poll_timeout_seconds."
+            )
+
     def _run_offchain_response_cycle(self) -> Generator:
         """Drive one off-chain response poll cycle and finish the behaviour.
 
@@ -623,6 +657,7 @@ class MechResponseBehaviour(MechInteractBaseBehaviour):
         integration only needs the FSM edge change, not a payload schema
         update.
         """
+        self._warn_if_round_timeout_below_poll_budget()
         with self.context.benchmark_tool.measure(self.behaviour_id).local():
             poller = OffchainResponsePoller(self)
             responses = yield from poller.run()
--- a/packages/valory/skills/mech_interact_abci/behaviours/response.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/response.py
@@ -38,7 +38,6 @@ from packages.valory.skills.mech_interact_abci.behaviours.request import V1_HEX_
 from packages.valory.skills.mech_interact_abci.models import MechResponseSpecs, Ox
 from packages.valory.skills.mech_interact_abci.payloads import JSONPayload
 from packages.valory.skills.mech_interact_abci.states.base import (
-    Event,
     MECH_RESPONSE,
     MechInteractionResponse,
     MechRequest,
@@ -54,9 +53,6 @@ BYTES32_HEX_FORMAT_SPEC = "064x"
 HEX_BASE = 16
 ADDRESS_ZERO = "0x0000000000000000000000000000000000000000"
 DELIVERY_MECH_INDEX = 1
-# Slack on top of the off-chain poll budget for payload assembly and
-# consensus, so the round timeout never fires mid-poll.
-OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS = 30.0
 
 
 class MechResponseBehaviour(MechInteractBaseBehaviour):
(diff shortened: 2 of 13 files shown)


bennyjo (approved):
Approving. Verified the flag-off FSM matches main (`MechResponseRound → ROUND_TIMEOUT → FinishedMechResponseTimeoutRound`) and that the runtime guard reads a real value — trader's composition writes `MechInteractEvent.ROUND_TIMEOUT` (1800s) into the same `event_to_timeout` the guard inspects, so consumer rebinds are visible and trader is well above the 330s budget. FSM and resilience framework audits ran clean.

One non-blocking note: the old class-level `RESPONSE_ROUND_TIMEOUT = 330.0` fallback *did* get copied by consumers' `chain()`, so it bounded the response round at 330s in composed apps even though the config rebind didn't take effect. Net effect of the revert for a consumer that enables `use_offchain` but leaves `ROUND_TIMEOUT` at the 30s default is 330s → 30s, downgraded from prevented to logged. No current offchain consumer sits at 30s (trader 1800s, market-resolver 1200s are strictly better off), so the guard's error log is now the load-bearing safety net for any future 30s-default offchain consumer — worth keeping in mind.

packages/valory/skills/mech_interact_abci/behaviours/response.py
```diff
@@ -53,6 +54,9 @@
 HEX_BASE = 16
 ADDRESS_ZERO = "0x0000000000000000000000000000000000000000"
 DELIVERY_MECH_INDEX = 1
+# Slack on top of the off-chain poll budget for payload assembly and
+# consensus, so the round timeout never fires mid-poll.
```
OjusWiZard: **Stale claim**: "so the round timeout never fires mid-poll" was true when this constant drove the `SharedState.setup` rebind, but this PR removes that enforcement — the constant now only computes the *warning* threshold, and the round timeout absolutely can still fire mid-poll (that's why the guard exists). Suggest rewording to something like: "used to compute the warning threshold in `_warn_if_round_timeout_below_poll_budget`".

packages/valory/skills/mech_interact_abci/behaviours/response.py
```diff
@@ -611,6 +615,36 @@ def _process_responses(  # pragma: no cover
                     f"There was an error in the mech's response: {self.current_mech_response.error}"
                 )
 
+    def _warn_if_round_timeout_below_poll_budget(self) -> None:
```
OjusWiZard: Nit: the method is named `_warn_...` but logs at `error` level (and the docstring says "Warn loudly" to bridge the gap). Either rename to `_error_if_...`/`_report_if_...` or log at `warning` — as-is, someone grepping logs by severity won't find this under "warn".

packages/valory/skills/mech_interact_abci/behaviours/response.py
```diff
@@ -611,6 +615,36 @@ def _process_responses(  # pragma: no cover
...
+        app instead. If it is below ``offchain_poll_timeout_seconds`` plus
+        overhead, the round times out mid-poll and every off-chain request
+        slower than the round timeout lands in
+        ``FinishedMechResponseTimeoutRound``.
+        """
+        poll_budget = (
+            float(self.mech_marketplace_config.offchain_poll_timeout_seconds)
+            + OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS
+        )
+        effective = self.context.state.round_sequence.abci_app.event_to_timeout.get(
```
OjusWiZard: **Type safety**: `self.context.state` resolves through `SkillContext.__getattr__` and is `Any` to mypy, so this entire chain (`.round_sequence.abci_app.event_to_timeout.get(...)`) is statically unchecked — a renamed framework attribute would only surface as a runtime `AttributeError`. The typed accessor already exists: `MechInteractBaseBehaviour.shared_state` (`behaviours/base.py:88-90`, `cast(SharedState, self.context.state)`), and `BaseSharedState.round_sequence -> RoundSequence` and `RoundSequence.abci_app -> AbciApp` are properly typed from there.

```suggestion
        effective = self.shared_state.round_sequence.abci_app.event_to_timeout.get(
```

packages/valory/skills/mech_interact_abci/behaviours/response.py
```diff
@@ -611,6 +615,36 @@ def _process_responses(  # pragma: no cover
...
+            Event.ROUND_TIMEOUT
+        )
+        if effective is not None and effective < poll_budget:
+            self.context.logger.error(
```
OjusWiZard: **Design question — log-only for a deterministic 100%-failure misconfiguration**: a consumer that enables `use_offchain` but keeps the 30s framework-default timeout (optimus, meme-ooorr per the PR description) will have *every* off-chain request time out mid-poll into `FinishedMechResponseTimeoutRound`, with this recurring error log as the only signal. The repo's precedent for config incoherence is fail-fast (`validate_configuration` raises `ValueError` at boot). If a flag-ON deployment with an unfit timeout is never legitimate, raising here would be strictly better than logging.

Related downstream consequence: when the round times out mid-poll, the behaviour is torn down before `finish_behaviour`, and nothing clears `offchain_pending_request` — the mech's eventual result is silently orphaned. In a misconfigured deployment that happens on effectively every request.

packages/valory/skills/mech_interact_abci/behaviours/response.py
```diff
@@ -623,6 +657,7 @@ def _run_offchain_response_cycle(self) -> Generator:
         integration only needs the FSM edge change, not a payload schema
         update.
         """
+        self._warn_if_round_timeout_below_poll_budget()
```
OjusWiZard: **Test gap**: all four `TestOffchainPollBudgetGuard` tests call `_warn_if_round_timeout_below_poll_budget()` directly; nothing exercises `_run_offchain_response_cycle`, so a refactor that drops this call-site line would leave the suite green — which defeats the purpose of the guard being the flag-ON safety net. Cheap fix: patch `OffchainResponsePoller`/`finish_behaviour`, drive this generator, and assert the guard was invoked.

packages/valory/skills/mech_interact_abci/tests/test_response_behaviour.py
```diff
@@ -330,3 +331,66 @@ def test_dispatches_to_legacy_when_v1(self, _mock: MagicMock) -> None:
...
+        """
+        behaviour = self._make_behaviour_with_timeout(
+            poll_timeout_seconds=300.0, effective_timeout=30.0
+        )
+        behaviour._warn_if_round_timeout_below_poll_budget()
+        behaviour.context.logger.error.assert_called_once()
+        log_msg = behaviour.context.logger.error.call_args[0][0]
+        assert "330.0" in log_msg  # poll budget = 300 + 30 overhead
+
+    def test_no_error_when_effective_timeout_equals_poll_budget(self) -> None:
```
OjusWiZard: Nit: this test and the two below it are structurally identical (build behaviour, call guard, `assert_not_called`), differing only in the two timeout values — a natural `@pytest.mark.parametrize`, which this suite already uses elsewhere (e.g. `test_models.py`). Would cut ~12 duplicated lines.

OjusWiZard (commented):
Multi-perspective review (code, comments, tests, silent failures, type design, simplification). The revert itself is clean: FSM spec verified with `autonomy analyse fsm-specs`, `autonomy packages lock --check` passes, no dangling `RESPONSE_ROUND_TIMEOUT` references, and the guard is correctly gated to the offchain path. Inline comments below cover the findings on the new guard code.

Advisories that don't anchor to diff lines:

- **`models.py:268`** — `offchain_poll_timeout_seconds` is annotated `float` but never coerced; both `response.py` (new) and `offchain_response.py:169` (pre-existing) defensively re-wrap it in `float(...)`. Coercing once in `MechMarketplaceConfig.__post_init__` (`object.__setattr__` since the dataclass is frozen) would make the annotation trustworthy and drop the duplicated casts.
- **Earlier signal** — the same read-only comparison could also run once at `SharedState.setup()` so a misconfiguration surfaces at startup instead of on the first off-chain round; worth verifying the init ordering first, since ordering is exactly what broke the removed rebind.
- **Docstring rot risk** — the guard docstring names trader's `mech_interact_round_timeout_seconds` attribute, which nothing in this repo can verify; it goes silently stale if trader renames it.

packages/valory/skills/mech_interact_abci/behaviours/response.py
DIvyaNautiyal07: Fixed. The constant moved to behaviours/base.py with the guard and the comment now says it computes the minimum round timeout enforced by `_check_round_timeout_fits_poll_budget`, no claim about the timeout not firing.

packages/valory/skills/mech_interact_abci/behaviours/response.py
DIvyaNautiyal07: Gone, the method is now `_check_round_timeout_fits_poll_budget` and it raises instead of logging, so the severity mismatch no longer exists.

packages/valory/skills/mech_interact_abci/behaviours/response.py
DIvyaNautiyal07: Done, the guard now goes through the typed `self.shared_state` accessor.

packages/valory/skills/mech_interact_abci/behaviours/response.py
DIvyaNautiyal07: Agreed and changed. The guard now raises ValueError, matching the `validate_configuration` precedent. It runs at the top of both `_run_offchain_request_cycle` and `_run_offchain_response_cycle`, so a misconfigured deployment fails before the request is posted or paid for, not just before the response poll. `async_act_wrapper` only catches StopIteration, so the ValueError propagates and the agent crash-loops, which is the operator signal we want for a deployment that can never work.

One correction on the downstream consequence: `offchain_pending_request` is not silently orphaned. `OffchainRequestExecutor.run()` (offchain_request.py:671-691) checks for a pending request first and resumes it by re-POSTing the same request_id with no new payment, so a mid-poll teardown picks the result up on the next cycle. In a misconfigured deployment that never got a chance to matter anyway, since the request round now raises before anything is sent.

packages/valory/skills/mech_interact_abci/behaviours/response.py
DIvyaNautiyal07: Added. Two integration tests drive the real `_run_offchain_response_cycle` generator: the misconfigured case asserts the ValueError propagates and `OffchainResponsePoller` is never constructed, the fit case runs a stubbed poller through to the payload. There is also a matching test for `_run_offchain_request_cycle` asserting the executor is never built, since that call site is the one protecting the payment.

packages/valory/skills/mech_interact_abci/tests/test_response_behaviour.py
DIvyaNautiyal07: Done, the three accept cases (equal, above, missing entry) are one parametrized test now.

packages/valory/skills/mech_interact_abci/behaviours/base.py
```diff
@@ -310,6 +318,37 @@ def wait_for_condition_with_sleep(  # pragma: no cover
...
+        """
+        poll_budget = (
+            self.mech_marketplace_config.offchain_poll_timeout_seconds
+            + OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS
+        )
+        effective = self.shared_state.round_sequence.abci_app.event_to_timeout.get(
+            Event.ROUND_TIMEOUT
+        )
+        if effective is not None and effective < poll_budget:
+            raise ValueError(
```
OjusWiZard, line 344: **What this raise actually does at runtime — confirm it's the intended blast radius.** Two independent traces through the pinned framework agree: `async_act_wrapper` only catches `StopIteration`; open-aea's `act_wrapper` wraps this as `AEAActException` and re-raises; and since neither this repo nor trader overrides `skill_exception_policy`, the default `propagate` policy sets the agent loop to error state and re-raises — **terminating the entire agent process, all composed skills included**. Because the guard runs on every entry to the round (not once at boot) and the misconfiguration is static, a supervised deployment (docker restart policy, k8s) crash-loops indefinitely.

For the **request-side** call this is exactly right: loud failure before any payment.

The **response-side** call has a sharper edge: a pending mech request legitimately survives a restart (Tendermint persists state), so a redeploy mid-flight that lowers the effective round timeout makes the response round raise and crash-loop the whole service over a request that was **already paid for**, with no recovery path except an operator config fix. The old log-only behavior contained that to one degraded interaction. Options if this matters: gate the response-side check on how the pending request was actually dispatched (persisted mode rather than live config), or keep the hard raise request-side only and degrade loudly response-side. If crash-looping on a mid-flight redeploy with broken config is acceptable collateral for the operational loudness, that's defensible — but it should be an explicit decision.

packages/valory/skills/mech_interact_abci/behaviours/base.py
```diff
@@ -310,6 +318,37 @@ def wait_for_condition_with_sleep(  # pragma: no cover
...
+                f"use_offchain is enabled but the effective mech-interact round timeout "
+                f"({effective}s) is below the off-chain poll budget "
+                f"(offchain_poll_timeout_seconds + {OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS}s "
+                f"overhead = {poll_budget}s). The response round would time out mid-poll "
```
OjusWiZard: **Message names the wrong round when raised from the request side.** This guard now also fires from `MechRequestRound` (`_run_offchain_request_cycle`), but the message hardcodes "The response round would time out mid-poll" — an operator reading the traceback from the request round sees text pointing at a different round. The docstring above has the same response-only framing ("The off-chain poll runs inside ``MechResponseRound``"). Suggest making both call-site-neutral, e.g. "the off-chain poll (run during the response round) would time out mid-poll" plus a docstring note that the check guards both offchain cycles.

packages/valory/skills/mech_interact_abci/tests/test_response_behaviour.py
```diff
@@ -330,3 +333,107 @@ def test_dispatches_to_legacy_when_v1(self, _mock: MagicMock) -> None:
...
+    def test_accepts_when_effective_timeout_fits_or_unknown(
+        self, effective_timeout: Optional[float]
+    ) -> None:
+        """A round timeout at/above the budget, or missing entirely, is accepted."""
+        behaviour = self._make_behaviour_with_timeout(
+            poll_timeout_seconds=300.0, effective_timeout=effective_timeout
+        )
+        behaviour._check_round_timeout_fits_poll_budget()
+
+    def test_offchain_response_cycle_raises_before_polling_on_misconfig(self) -> None:
```
OjusWiZard, line 407: **No flag-off tripwire for the response behaviour.** Nothing in the suite calls `MechResponseBehaviour.async_act` at all (the request side has `test_async_act_skips_when_priority_mech_empty`; the response side has nothing). The guard is safely nested behind `use_offchain is True` today, but a future edit that hoists `_check_round_timeout_fits_poll_budget` up into `async_act` — or breaks the gate — would violate the epic's core invariant (flag-off byte-identical to main) with a fully green suite. A minimal `async_act` test with `use_offchain=False` asserting the guard is never invoked would close this.

packages/valory/skills/mech_interact_abci/tests/test_request_behaviour.py
```diff
@@ -532,3 +534,31 @@ def test_successful_selection_clears_stale_failure_reason(
 
         assert result == "0xgood"
         assert mock_shared.last_failure_reason is None
+
+
+class TestOffchainRequestCycleGuard:
+    """Tests for the poll-budget guard at the off-chain request cycle entry."""
+
+    def test_offchain_request_cycle_raises_before_executor_on_misconfig(self) -> None:
```
OjusWiZard, line 545: **Asymmetry with the response side**: the response cycle got a "completes through to the payload" happy-path test, but the request cycle only got this raise test. `_payload_from_offchain_result`'s field mapping (~9 fields incl. `tx_submitter`, `offchain_pending_request`, `offchain_last_failure_reason`) is not exercised end-to-end by any test — the executor's internals are unit-tested in `test_offchain_request_helpers.py`, but not the cycle wrapper's payload construction. A mirror of `test_offchain_response_cycle_completes_when_timeout_fits` with a stub executor would cover it.

packages/valory/skills/mech_interact_abci/tests/test_response_behaviour.py
```diff
@@ -330,3 +333,107 @@ def test_dispatches_to_legacy_when_v1(self, _mock: MagicMock) -> None:
 
         result = behaviour._check_match(pending, request, is_first_pending=True)
         assert result is True
+
+
+class _EmptyResponsesPoller:
```
OjusWiZard, line 338: Nit: this stub replaces `OffchainResponsePoller` wholesale with no `spec=` link to the real class, so if the real poller's `__init__`/`run()` signature or yield/return contract drifts, this test keeps passing against broken wiring. Consider `MagicMock(spec=OffchainResponsePoller)` for construction or a brief comment accepting the drift risk.

packages/valory/skills/mech_interact_abci/models.py
```diff
@@ -273,6 +272,19 @@ class MechMarketplaceConfig:
 
     def __post_init__(self) -> None:
         """Validate configuration after initialization."""
+        # Coerce so the float annotations hold even when the values arrive
```
OjusWiZard, line 275: Nit: the comment explains why the coercion exists but not why `object.__setattr__` — worth one more clause ("via ``object.__setattr__`` because the dataclass is frozen") for readers unfamiliar with the frozen-dataclass `__post_init__` idiom.

packages/valory/skills/mech_interact_abci/tests/test_models.py
```diff
@@ -203,6 +203,24 @@ def test_invalid_offchain_timing_params_raise(
                 **{field: value},  # type: ignore[arg-type]
             )
 
+    def test_offchain_poll_values_coerced_to_float(self) -> None:
```
OjusWiZard: Nit: only int inputs are covered. A quoted YAML override delivers a string, and `float("300")` succeeding is precisely what makes that shape work now (previously `"300" <= 0` raised `TypeError` at boot). One extra parametrized case with `"300"` would pin that contract.

OjusWiZard (commented):
Round-2 review of 5d3d58dd (same multi-perspective stack as round 1, scoped to the delta). The fix commit is mechanically sound: fingerprints and the package hash re-verified programmatically (`check_fingerprint` passes, IPFS hash matches `packages.json`), no import cycle from the new `Event` import in `behaviours/base.py`, the guard is correctly gated to the offchain branches only, and the float coercion ordering in `__post_init__` is correct — it actually *fixes* a boot crash for quoted-numeric YAML overrides (`"300" <= 0` used to raise `TypeError`).

The inline comments below are the round-2 findings. The main one (crash-loop semantics) is a design decision to confirm, not a defect: two independent traces through the vendored framework agree on what the raise does at runtime.

packages/valory/skills/mech_interact_abci/behaviours/base.py
DIvyaNautiyal07: Keeping the raise on both sides, this is intended. A below-budget timeout is never a legitimate deployment, so the crash-loop is the loudest operator-visible signal we have, same posture as the validate_configuration precedent. On the paid-pending edge: the pending request lives in synchronized data, so it survives the restart via Tendermint state, and OffchainRequestExecutor.run() resumes it by re-POSTing the same request_id with no new payment once the timeout is corrected. So the blast radius of a mid-flight redeploy is downtime until the config is fixed, not lost funds. Degrading response-side instead would keep the service half-alive with every off-chain request timing out mid-poll, which is easier to miss than a crash-loop.

packages/valory/skills/mech_interact_abci/behaviours/base.py
DIvyaNautiyal07: Done in c92e458, reworded the docstring and message to be call-site-neutral. The docstring now names both rounds and the message says "the round running the off-chain cycle".

packages/valory/skills/mech_interact_abci/tests/test_response_behaviour.py
DIvyaNautiyal07: Added test_flag_off_async_act_never_invokes_guard in c92e458. It drives MechResponseBehaviour.async_act with use_offchain=False and asserts the guard is never called and the on-chain payload path completes.

packages/valory/skills/mech_interact_abci/tests/test_request_behaviour.py
DIvyaNautiyal07: Added test_offchain_request_cycle_lifts_executor_result_onto_payload in c92e458, mirroring the response-side completion test. The executor is stubbed at its boundary with a canned OffchainCycleResult, _payload_from_offchain_result runs for real, and every payload field is asserted one-to-one with distinct sentinels so a swapped or dropped kwarg fails.

packages/valory/skills/mech_interact_abci/tests/test_response_behaviour.py
DIvyaNautiyal07: Kept the named stub since MagicMock(spec=) gets awkward for a generator-returning run(), and added a drift-risk comment in c92e458 tying its constructor and run() signature to OffchainResponsePoller.

packages/valory/skills/mech_interact_abci/models.py
DIvyaNautiyal07: Done in c92e458, the comment now notes the writes go via object.__setattr__ because the dataclass is frozen.

packages/valory/skills/mech_interact_abci/tests/test_models.py
DIvyaNautiyal07: Done in c92e458, the coercion test is now parametrized with quoted-string values ("5"/"300"), pinning the fix for the boot crash from quoted YAML overrides.

packages/valory/skills/mech_interact_abci/behaviours/base.py
```diff
@@ -310,6 +318,39 @@ def wait_for_condition_with_sleep(  # pragma: no cover
...
+        )
+        effective = self.shared_state.round_sequence.abci_app.event_to_timeout.get(
+            Event.ROUND_TIMEOUT
+        )
+        if effective is not None and effective < poll_budget:
+            raise ValueError(
+                f"use_offchain is enabled but the effective mech-interact round timeout "
+                f"({effective}s) is below the off-chain poll budget "
+                f"(offchain_poll_timeout_seconds + {OFFCHAIN_POLL_TIMEOUT_OVERHEAD_SECONDS}s "
+                f"overhead = {poll_budget}s). The round running the off-chain cycle "
```
OjusWiZard, line 348: **The call-site-neutral rewording now overclaims for the request side.** The docstring says "every off-chain cycle times out mid-poll" and this message says "The round running the off-chain cycle would time out mid-poll on every request" — but the request cycle never polls: `OffchainRequestExecutor.run()` POSTs once per ranked mech (bounded by `offchain_failover_max_retries`) and never consumes `offchain_poll_timeout_seconds`, so there is no mid-poll state it can time out in. Only `OffchainResponsePoller` polls. The request-side guard's real rationale is the one already stated at its call site: fail before any payment, because the *response* round that follows would be doomed.

Suggested direction: keep the shared "bounded by the composed app's ``ROUND_TIMEOUT``, read off the live app" framing, scope the mid-poll claim to the response round, and reference fail-before-payment for the request round. Non-blocking — the guard's logic and placement are correct; only the stated mechanism is off for one of the two call sites.

OjusWiZard (approved):
Round-3 review of c92e458f: approving. The delta is clean — fingerprints and package hash re-verified against freshly regenerated values, black clean, and all three new tests were **mutation-tested**: hoisting the guard into `async_act` trips the flag-off tripwire, swapping payload kwargs fails the field-by-field request test, and the quoted-string case genuinely reproduces the pre-coercion boot crash. Patch hygiene (instance-level guard mock, class-property `synchronized_data` patch) verified with no cross-test pollution.

One wording fix left as an inline comment below — non-blocking.

bennyjo (approved):
Re-reviewed at c92e458. The substance has changed materially since my earlier approval (log → hard `ValueError`, float coercion, closed test gaps), and my concern is resolved: the flag-ON safety net now *prevents* an unfit-timeout off-chain deployment rather than logging it, so the hypothetical 30s-default consumer is back to prevented. Flag-off FSM is byte-identical to main, no live consumer is affected (trader 1800s, market-resolver 1200s > 330s budget), and the crash-loop semantics are an explicit, defensible decision (downtime-until-fixed, not lost funds — pending requests resume via re-POST with no new payment).

One non-blocking nit remains open (base.py:344–351): the guard's message/docstring claim "mid-poll on every request" even when raised from the request round, which never polls — `OffchainRequestExecutor.run()` POSTs once per ranked mech and never touches `offchain_poll_timeout_seconds`. The guard's logic and placement are correct on both sides; only the stated mechanism is off for the request call site. Reasonable to fold in here or defer. Approving.

PR #102, opened 2026-07-09
fix(offchain): carry request_context and schema_version in ipfs_data

DIvyaNautiyal07:
## Summary

- Off-chain `build_request_metadata` now emits `request_context` and `schema_version` unconditionally, matching the on-chain `_send_metadata_to_ipfs` payload shape. Previously the off-chain builder rebuilt the metadata dict by hand and silently dropped both fields, so analytics ETL that reads `raw_content->request_context->market_id`/`market_prob` lost attribution for any request that went through the off-chain path.
- Mirrors the on-chain clobber warning: when `extra_attributes` overlap the reserved keys (`prompt`, `tool`, `nonce`, `schema_version`, `request_context`), the builder now emits a warning instead of silently overwriting.
- Adds a drift-guard parity test that feeds the same `MechMetadata` through the on-chain path (`asdict` + merge extras) and the off-chain builder and asserts the resulting dicts are equal.

## Design doc

`docs/offchain_request_context_fix.md` (in this branch's parent) has the full context, invariants, and downstream ripple table.

## Test plan
- [x] `uv run pytest packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py` — 94 pass, including the new `TestBuildRequestMetadata` cases and the two parity tests.
- [x] `uv run tomte tox -e py3.12-darwin` — full skill+contract suite, 598 pass.
- [x] `uv run tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — all green.
- [x] `uv run tomte tox -p -e check-packages -e check-hash -e check-third-party-hashes -e check-abci-docstrings -e check-abciapp-specs -e check-handlers -e check-dependencies -e check-doc-hashes -e gitleaks -e safety -e bandit` — all green.
- [x] `autonomy push-all` — new `mech_interact_abci` hash `bafybeigoqw3b3pbjh5yvicshlttum64ekmoffluqyw3ulzyj45iwygee6i` published.

### DIFF
--- a/packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
@@ -90,6 +90,8 @@ from packages.valory.skills.mech_interact_abci.states.base import (
     OFFCHAIN_503_ALL_MECHS,
     OFFCHAIN_BAD_RESPONSE,
     OFFCHAIN_TIMEOUT_ALL_MECHS,
+    SCHEMA_VERSION,
+    merge_extra_attributes,
 )
 from packages.valory.skills.mech_interact_abci.states.request import (
     OFFCHAIN_DEPOSIT_TX_SUBMITTER,
@@ -186,7 +188,8 @@ def build_request_metadata(
     extra_attributes: Optional[Dict[str, Any]] = None,
     nonce_str: Optional[str] = None,
     request_context: Optional[Dict[str, Any]] = None,
-    schema_version: str = "2.0",
+    schema_version: str = SCHEMA_VERSION,
+    logger: Optional[logging.Logger] = None,
 ) -> Tuple[str, str, str]:
     """Build the offchain request metadata and its on-chain hash.
 
@@ -201,6 +204,12 @@ def build_request_metadata(
     ``truncated_hash`` is the ``0x`` + 62-hex form the on-chain
     commitment uses and ``ipfs_data`` is the JSON string carried as
     the ``ipfs_data`` form field.
+
+    :param logger: Logger for the clobber warning. Callers on the executor
+        path pass ``self._logger`` so the warning routes through the AEA
+        logging pipeline; passing ``None`` (the module-level default) sends
+        the warning through the module logger instead — used by tests and
+        ad-hoc call sites.
     """
     metadata: Dict[str, Any] = {
         "prompt": prompt,
@@ -209,13 +218,7 @@ def build_request_metadata(
         "schema_version": schema_version,
         "request_context": request_context,
     }
-    if extra_attributes:
-        clobbered = extra_attributes.keys() & metadata.keys()
-        if clobbered:
-            _LOGGER.warning(
-                "extra_attributes override reserved request keys: %s", clobbered
-            )
-        metadata.update(extra_attributes)
+    merge_extra_attributes(metadata, extra_attributes, logger or _LOGGER)
     ipfs_data = json.dumps(metadata)
     cid_bytes = compute_cidv1_bytes(ipfs_data.encode("utf-8"))
     v1_file_hash_hex = "f" + cid_bytes.hex()
@@ -729,6 +732,7 @@ class OffchainRequestExecutor:
             nonce_str=request_meta.nonce,
             request_context=request_meta.request_context,
             schema_version=request_meta.schema_version,
+            logger=self._logger,
         )
 
         chain_id_int = yield from self._resolve_chain_id_int()
--- a/packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
+++ b/packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
@@ -53,6 +53,7 @@ import dataclasses
 import enum
 import hashlib
 import json
+import logging
 import re
 import uuid
 from dataclasses import dataclass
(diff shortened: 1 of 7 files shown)


OjusWiZard (approved):
Approving. Clean, correctly scoped parity fix — the offchain `build_request_metadata` now emits `schema_version` and `request_context` in the same shape/order as the on-chain `_send_metadata_to_ipfs` payload, so the analytics lake sees one request schema regardless of transport.

Verified:
- Key ordering and clobber-warning behaviour match the on-chain twin.
- `request_context=None` serialises to JSON `null`, matching `asdict(MechMetadata)`.
- `test_offchain_matches_onchain_payload_shape` is a solid drift guard (both populated + defaults).
- CI green across all checks, including `copyright_and_dependencies_check` (liccheck resolves cffi ≥2.1 fresh, so the `>=2.1,<3` allow-list matches).

Two non-blocking notes for later:
1. The new clobber warning uses a module-level `logging.getLogger(__name__)` rather than the AEA `context.logger`, so it may not surface through the agent's configured handlers in production.
2. `tox.ini` comment references cffi 2.1.0 (MIT-0) but `uv.lock` pins 2.0.0 — audited vs shipped version differ; consider relocking cffi to 2.1.x.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -182,21 +185,36 @@ def build_request_metadata(
...
         "prompt": prompt,
         "tool": tool,
         "nonce": nonce_str if nonce_str is not None else str(uuid.uuid4()),
+        "schema_version": schema_version,
+        "request_context": request_context,
     }
     if extra_attributes:
+        clobbered = extra_attributes.keys() & metadata.keys()
+        if clobbered:
+            _LOGGER.warning(
```
bennyjo: This warning uses a module-level logger, so it may not show up in the agent's logs. Could `_fresh_cycle` pass in `self._logger` instead?

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -182,21 +185,36 @@ def build_request_metadata(
     tool: str,
     extra_attributes: Optional[Dict[str, Any]] = None,
     nonce_str: Optional[str] = None,
+    request_context: Optional[Dict[str, Any]] = None,
+    schema_version: str = "2.0",
```
bennyjo: This "2.0" duplicates the default on `MechMetadata` — if one gets bumped later, the other will drift. Suggest reading the default from `MechMetadata` instead.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
```diff
@@ -157,6 +163,118 @@ def test_default_nonce_is_uuid(self) -> None:
...
+        warnings = [r for r in caplog.records if r.levelno == logging.WARNING]
+        assert any("extra_attributes override reserved" in r.message for r in warnings)
+        # Both clobbered keys are reported.
+        collision_msg = next(
+            r.message for r in warnings if "reserved request keys" in r.message
+        )
+        assert "prompt" in collision_msg
+        assert "request_context" in collision_msg
+
+    def test_offchain_matches_onchain_payload_shape(self) -> None:
```
bennyjo, line 260: This test copies the on-chain merge logic by hand, so if `request.py` changes, the test won't notice. A small shared helper used by both paths would guarantee they stay in sync — fine as a follow-up.

bennyjo (commented):
A few suggestion-level comments — nothing blocking, overall LGTM.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Done in 763bb4f. `build_request_metadata` now takes an optional `logger` kwarg and `_fresh_cycle` passes `self._logger`, so the clobber warning routes through the AEA logging pipeline. The module logger stays as a fallback for tests and ad-hoc call sites (kept `logger or _LOGGER` so the fallback is explicit at the call site). Also added a test that fails if the caller-supplied logger is not the one emitting the record.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Done in 763bb4f. Extracted `SCHEMA_VERSION = "2.0"` in `states/base.py` next to `MechMetadata`. Both `MechMetadata.schema_version` and `build_request_metadata` default reference the same constant, so a bump lands in one place.

packages/valory/skills/mech_interact_abci/tests/behaviours/test_offchain_request_helpers.py
DIvyaNautiyal07: Pulled this into 763bb4f rather than deferring. Added `merge_extra_attributes(payload, extras, logger)` in `states/base.py` — it pops the wrapper key, computes the clobbered set, and logs the warning (routed through the caller-supplied logger). Both `_send_metadata_to_ipfs` (on-chain), `build_request_metadata` (off-chain), and the parity test now go through it, so any future change to the merge/clobber semantics is picked up by all three call sites at once.

bennyjo (approved):
Approving — all three of my earlier comments were addressed in 763bb4f (logger routing with a discriminating test, shared SCHEMA_VERSION constant, and the merge logic extracted into merge_extra_attributes shared by both request paths and the parity test). CI green, package hashes consistent.

PR #103, opened 2026-07-10
feat(mech-interact): offchain epic — request/response behaviours (stack)

DIvyaNautiyal07:
## Summary

Stack rollup of the mech-interact offchain request/response work into `main`.
`main` currently has no offchain path; this PR fast-forwards it to the epic
head so the offchain request/response behaviours (used by trader / optimus /
market-resolver to talk to an offchain mech via the marketplace API) land in
one reviewable unit.

## Sub-PRs merged into `feat/offchain-epic`

- [#95](https://github.com/valory-xyz/mech-interact/pull/95) — feat(offchain): implement off-chain request/response behaviours
  Initial cut. Adds the offchain-request and offchain-response behaviours,
  the FSM anchors, the payment-type constants, and the skill wiring.
- [#96](https://github.com/valory-xyz/mech-interact/pull/96) — fix(offchain): 3 executor bugs from #95 uncovered by local e2e
  Follow-up on #95 after a live-mech end-to-end run: chain-id threading on
  executor contract reads, `_pick_next_mech` gating on
  `use_dynamic_mech_selection`, and single-call deposit via inline
  `GnosisSafe.get_raw_safe_transaction_hash`.
- [#98](https://github.com/valory-xyz/mech-interact/pull/98) — fix(offchain): extract inner `result` field to match on-chain response shape
  Aligns offchain-response payload shape with the on-chain path so
  downstream skills consume both paths uniformly.
- [#99](https://github.com/valory-xyz/mech-interact/pull/99) — refactor: consume canonical BalanceTracker + MechMarketplace, bump OA 0.21.26
  Switches to the canonical BalanceTracker + MechMarketplace contracts,
  aligns third-party OA hashes to v0.21.26, registers the mech upstream,
  and pins mech v0.34.2-rc2.
- [#101](https://github.com/valory-xyz/mech-interact/pull/101) — fix(offchain): revert RESPONSE_ROUND_TIMEOUT to shared ROUND_TIMEOUT
  Drops the dedicated response-round timeout in favour of the shared
  `ROUND_TIMEOUT` and fails fast when the poll budget can't fit inside it.
- [#102](https://github.com/valory-xyz/mech-interact/pull/102) — fix(offchain): carry `request_context` and `schema_version` in `ipfs_data`
  Threads `request_context` and `schema_version` through the offchain
  ipfs payload so the request-side and response-side match the on-chain
  parity contract.

## Post-merge

Downstream consumers (trader / optimus / market-resolver) that pin
`mech_interact_abci` will need a hash bump to pick up the offchain
behaviours. That's a follow-up per repo.

## Test plan

- [x] Sub-PRs individually reviewed and green on their own CI
- [ ] Full CI green on this rollup
- [ ] Downstream hash-bump PRs land cleanly once this merges


packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+                "request_id": request_id,
+                "delivery_rate": delivery_rate,
+                "nonce": nonce,
+                "ipfs_data": ipfs_data,
+            }
+        ).encode("utf-8")
+
+    # ---------- signing -----------------------------------------------------
+
+    def _sign_request_id(
```
OjusWiZard, line 1179: **[Critical] Signature likely can't satisfy a multi-owner Safe's on-chain verification.**

`_sign_request_id` signs with the individual agent's AEA key (raw ECDSA via `get_signature(is_deprecated_mode=True)`), but the `requester`/`sender` on the request is the **Safe address** (a contract). `MechMarketplace._verifySignedHash` branches on `requester.code.length > 0` and, for contract requesters, requires ERC-1271 (`ISignatureValidator(requester).isValidSignature(...)` → Gnosis Safe `checkSignatures`). A single 65-byte agent-key signature only passes `checkSignatures` when the Safe threshold is 1.

For a threshold > 1 service (the standard multi-agent OA shape): the mech's HTTP intake accepts the POST (no synchronous sig check), the FSM sees `OFFCHAIN_DONE`, but on-chain settlement reverts with `SignatureNotValidated` — the response poller then burns the full `offchain_poll_timeout_seconds` and times out. No test exercises threshold > 1, and none of the C2–C6 review notes mention EIP-1271.

Can you confirm what Safe threshold the #96 live e2e used? If threshold > 1 is a supported deployment, the signing scheme needs a real fix (aggregated Safe signature bundle collected via consensus, or an alternate authorization path on the settlement side).

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,380 @@
...
+
+    @property
+    def _synced(self) -> Any:
+        return self._b.synchronized_data
+
+    @property
+    def _logger(self) -> Any:
+        return self._b.context.logger
+
+    def run(self) -> Generator[None, None, List[MechInteractionResponse]]:
```
OjusWiZard, line 104: **[Critical — test gap] `run()` and its matching logic have zero real test coverage.**

`run`, `_find_target_response`, `_apply_snapshot`, and `serialise_responses` are never invoked by any test. `test_offchain_response_poller.py` only drives `_poll_until_terminal`/`_load_pending` directly, and the wiring test in `test_response_behaviour.py` (`TestOffchainPollBudgetGuard`) substitutes a hand-written stub whose own docstring admits nothing links it mechanically to this class.

This is the code that matches a delivered mech result to the correct pre-populated `MechInteractionResponse` (by `requestId`/`requestIds`), falls back to appending a placeholder on match-miss, and mutates the target in place — a bug here (e.g. an `int(request_id, 16)` mismatch or broken in-place mutation after a refactor) would silently drop or duplicate delivered answers with no regression signal.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+            continue
+
+        return OffchainCycleResult(
+            offchain_result=Event.OFFCHAIN_ALL_FAILED.value,
+            last_failure_reason=last_failure or OFFCHAIN_TIMEOUT_ALL_MECHS,
+        )
+
+    # ---------- resume after deposit ---------------------------------------
+
+    def _retry_pending(
```
OjusWiZard, line 928: **[Important] `_retry_pending` mislabels every non-DONE outcome as "insufficient balance" and logs nothing.**

The retry-after-deposit POST can come back `DEPOSIT_NEEDED`, `OVER_CAP`, `TIMEOUT`, `SERVER_BUSY`, or `BAD_RESPONSE`, but every non-`DONE` outcome is hard-coded to `last_failure_reason=OFFCHAIN_402_INSUFFICIENT`, and — unlike `_fresh_cycle`, which logs on every branch — this function contains no logging at all (`attempt.status_code`/`attempt.outcome` are never recorded).

Scenario: the deposit settles, then the retry POST hits a network blip or a 503. The operator sees `OFFCHAIN_ALL_FAILED` + `offchain_402_insufficient` and investigates deposit sizing instead of network/server health. Only the genuine second-402 case is tested (`test_402_on_retry_surfaces_insufficient`).

Suggestion: log outcome/status unconditionally in the failure branch and map `last_failure_reason` via `_failure_label_for` instead of hard-coding.

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -497,6 +544,55 @@ def is_marketplace_v2(self) -> Optional[bool]:
...
+        One of ``offchain_done``, ``offchain_deposit_needed``,
+        ``offchain_all_failed``, or ``None`` when the on-chain path ran.
+        ``MechRequestRound.end_block`` reads this to dispatch to the right
+        ``Event`` variant. Persisted as a string so the value flows through
+        the standard payload/selection-key path.
+        """
+        return cast(Optional[str], self.db.get("offchain_result", None))
+
+    @property
+    def offchain_pending_request(self) -> Optional[Dict[str, Any]]:
```
OjusWiZard, line 583: **[Important] Silently swallows corruption + docstring understates when this is populated.**

1. This property collapses JSON-decode/shape failures to `None` with no logging. Since the payload field is always a JSON string (`PendingRequest.to_json()`), the `json.loads` path here is the *primary* path, not an edge case — and a corrupted value becomes indistinguishable from "nothing pending." That defeats the corruption warnings both consumers were hardened with: `_load_pending_request`'s `if not raw` guard fires first and returns silently. Scenario: a 402 deposit lands on-chain, the persisted string is corrupted after a restart, and the agent silently runs `_fresh_cycle()` against a possibly different mech — paid-for correlation lost with no diagnostic trail. Suggest logging a warning here when the key is present but unparseable.

2. Docstring says "populated when the previous attempt returned a structured 402," but `_fresh_cycle`'s DONE branch also sets `pending_request_json=pending.to_json()` on every successful request — and that's what `OffchainResponsePoller._load_pending` reads to get `mech_url`/`request_id` to poll. Worth documenting both roles.

3. No test exercises this accessor's real string-JSON path — every executor/poller test injects a raw dict via a stub.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+            numRequests=1,
+            nonce=metadata_nonce,
+            result=None,
+            error="Unknown",
+        )
+        return json.dumps([dataclasses.asdict(placeholder)], ensure_ascii=True)
+
+    # ---------- state-on-the-wire helpers ----------------------------------
+
+    def _load_pending_request(self) -> Optional[PendingRequest]:
```
OjusWiZard, line 1854: **[Important] Copy-paste drift vs `OffchainResponsePoller._load_pending`.**

This helper and `offchain_response.py:347` do the same job (read `offchain_pending_request`, guard, `PendingRequest.from_dict`, log-and-discard on failure) but with divergent guards: here `if not raw` treats `{}` as silently "nothing pending," while the poller's `isinstance(raw, dict)` treats `{}` as corrupt-and-warn. No comment explains the difference, so it reads as drift rather than intent.

Suggest consolidating into one shared `load_pending_request(synced, logger)` next to `PendingRequest` — picking the isinstance-based guard deliberately (it's stricter and matches the `_load_pending` docstring's stated contract).

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,380 @@
...
+            mech_url=pending.mech_url,
+            request_id_int_str=str(int(pending.request_id, 16)),
+        )
+
+        self._apply_snapshot(target, snapshot)
+        return responses
+
+    # ---------- polling loop -----------------------------------------------
+
+    def _poll_until_terminal(
```
OjusWiZard, line 148: **[Important] Per-agent HTTP polling can diverge inside a `CollectSameUntilThresholdRound`.**

Each agent independently GETs `/fetch_offchain_info` on its own wall clock (`time.monotonic()` deadline, `_MAX_CONSECUTIVE_5XX` counter). Unlike the on-chain event-log polling this replaces — where blockchain state converges identically for every reader — a real HTTP server's behavior is per-connection: a 503 burst hitting one agent's path, or one agent crossing its deadline moments before the mech flips to "ok," yields different payloads for the same `request_id` in the same round → recurring `NO_MAJORITY`, which loops the round and re-polls from scratch. An agent with persistently worse connectivity to the mech could stall the round indefinitely.

Not necessarily a blocker, but worth a deliberate design note (and possibly a bounded number of NO_MAJORITY loops before surfacing a failure event).

packages/valory/skills/mech_interact_abci/states/request.py
```diff
@@ -63,6 +92,17 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
         if event != Event.DONE:
             return res
 
+        # Offchain dispatch. ``MechRequestBehaviour`` sets ``offchain_result``
+        # on the payload when ``use_offchain=True``; the value drives which
+        # ``OFFCHAIN_*`` event leaves the round so the FSM routes to the
+        # right final state. ``None`` means the on-chain path ran today's
+        # behaviour and the existing ``DONE`` / ``SKIP_REQUEST`` rules apply.
+        offchain_event = _OFFCHAIN_RESULT_TO_EVENT.get(
```
OjusWiZard: **[Important] Unrecognized `offchain_result` values are silently reinterpreted as on-chain outcomes.**

`_OFFCHAIN_RESULT_TO_EVENT.get(...)` returns `None` for any unknown value and `end_block` treats `None` as "the on-chain path ran," falling through to the `SKIP_REQUEST`/`DONE` rules. `offchain_result` is a plain `str` tied to the `Event` enum only by convention at the three construction sites — a future outcome added to the executor but not to this map produces a wrong FSM transition with no error signal.

Related test gap: nothing asserts `MechInteractAbciApp.transition_function[MechRequestRound][Event.OFFCHAIN_*]` routes to the correct final rounds — `test_rounds.py` covers `end_block`'s return values but not the transition table (contrast `TestResponseRoundTimeoutEvent` in `test_models.py`, which does check the table).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+class OffchainRequestExecutor:
+    """Coordinator for one ``MechRequestRound`` offchain cycle.
+
+    Holds a reference back to the parent behaviour so it can yield through
+    the framework's HTTP / signing / contract API helpers. Stateless across
+    cycles — persistent state lives on ``SynchronizedData`` so it survives
+    the deposit-then-retry round-trip.
+    """
+
+    def __init__(self, behaviour: "Any") -> None:
```
OjusWiZard, line 672: **[Important] `Any`-typed collaborator erases mypy coverage across ~2,200 lines (and is duplicated).**

`self._b`, `_config`, `_synced`, and `_logger` are all `Any`, so with `--disallow-untyped-defs` in CI, virtually every attribute access in this module and `offchain_response.py` is invisible to the type checker. A rename on `MechParams`/`MechMarketplaceConfig`/`SynchronizedData` surfaces as a runtime `AttributeError` mid-cycle — potentially after an HTTP POST or deposit tx has been built — instead of a CI failure.

The same `__init__` + three properties are byte-for-byte duplicated in `OffchainResponsePoller` (`offchain_response.py:88-102`). Extracting a small typed shared base/mixin fixes both the duplication and the type-erasure in one change.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+            if attempt.outcome is OffchainAttemptOutcome.DEPOSIT_NEEDED:
+                # Defence against a malicious or compromised mech directing
+                # the auto-deposit to an attacker-chosen address: cross-check
+                # ``challenge.pay_to`` (and ``asset`` on the token path)
+                # against the canonical BalanceTracker the marketplace has
+                # registered for the mech's ``paymentType``. A mismatch is
+                # treated as a misbehaving mech (BAD_RESPONSE) and failover
+                # moves on rather than landing a redirected deposit.
+                destination_ok = yield from self._validate_402_destination(
+                    challenge=attempt.challenge,  # type: ignore[arg-type]
```
OjusWiZard, line 859: **[Important] `Optional[PaymentChallenge]` narrowed via `# type: ignore` instead of the type system.**

The invariant "`challenge` is non-None iff `outcome` is `DEPOSIT_NEEDED`/`OVER_CAP`" holds only by convention in `_post_signed_request`, and these three `type: ignore[arg-type]`/`[union-attr]` sites (lines 842, 850, 890) bypass mypy rather than narrowing. A future edit that returns those outcomes without setting `challenge` becomes an `AttributeError` deep in the deposit-building path — after destination validation may have already run — instead of a review-time mypy failure.

Suggest either an explicit assert-narrow, or modeling the outcome variants structurally (small tagged union / per-outcome dataclass carrying a non-Optional challenge).

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -87,14 +98,44 @@ class MechMetadata:
...
+    :param extras: Extra tool parameters. ``None`` or empty is a no-op.
+    :param logger: Logger to route the clobber warning through. Callers
+        that want the warning surfaced in the agent's log pipeline pass
+        their behaviour's ``context.logger`` / ``_logger``; passing
+        ``None`` suppresses the warning entirely.
+    :return: The mutated ``payload`` (returned for chainability).
+    """
+    if not extras:
+        return payload
+    clobbered = extras.keys() & payload.keys()
```
OjusWiZard, line 148: **[Important] Warn-and-clobber of protocol-reserved keys permits silent tool/parameter substitution.**

`extra_attributes` is a free-form, unvalidated dict, and this helper lets it overwrite `prompt`, `tool`, `nonce`, `schema_version`, and `request_context` with only a log warning before `payload.update(extras)`. Since this feeds the mech's actual instruction payload on both paths — and on the offchain path now also drives a real HTTP POST plus (on 402) a real Safe deposit — a clobbered `tool` means the service pays for and believes it requested tool X while the mech executes something else.

If warn-and-merge was a deliberate #102 review outcome, consider at least hard-rejecting the reserved-key subset (`prompt`/`tool`/`nonce`/`schema_version`/`request_context`) while still allowing benign extras.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+        # ``priority_mech_address`` that has not already been tried.
+        fallback = self._config.offchain_url
+        static_address = self._config.priority_mech_address
+        if fallback and static_address and static_address.lower() not in attempted_lc:
+            return static_address, fallback
+        return None, None
+
+    # ---------- deposit multisend ------------------------------------------
+
+    def _build_deposit_tx(
```
OjusWiZard, line 1532: **[Important — test gaps] Several executor branches never driven end-to-end:**

- `_build_deposit_tx`'s NVM short-circuit (`_PAYMENT_NVM_NATIVE`/`_PAYMENT_NVM_TOKEN`) and the unknown-label `return None` — no test drives a 402 against an NVM-classified mech through `_fresh_cycle` to confirm it surfaces `OFFCHAIN_402_INSUFFICIENT` instead of attempting a deposit that reverts by design.
- `_pick_next_mech`: the per-mech `mech.http_url or self._config.offchain_url` fallback inside the ranked loop, and the `use_dynamic_mech_selection=False` gate with a non-empty ranked list (added specifically for hardcoded-priority-address consumers) — all existing tests use `ranked_mechs=[]` or mechs with real URLs.
- `_resolve_payment_type` wrong-length/type branches and `_resolve_chain_id_int`/`_read_on_chain_nonce` non-numeric-value except-branches (STATE performative with bad value) — only the non-STATE failure is covered.
- `OffchainRequestExecutor.run()`'s fresh-vs-resume dispatch is never invoked anywhere; only its two callees are driven directly.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+    OVER_CAP = "over_cap"
+    """The 402 shortfall exceeded ``auto_deposit_cap_per_cycle``; surface to consumer."""
+
+    TIMEOUT = "timeout"
+    """The HTTP call raised (framework error) or returned ``None``.
+
+    Counts toward the failover budget. The framework helper used here
+    (``get_http_response``) does not expose a per-call timeout argument,
+    so this is the framework's own raise/None signal rather than a
+    deadline policed locally — see the ``offchain_http_timeout_seconds``
```
OjusWiZard: **[Docs] Reference to a config that doesn't exist.** `offchain_http_timeout_seconds` appears nowhere in the codebase (the real knob is `offchain_poll_timeout_seconds` in `models.py`), and there are no "module-level review notes" in this file to consult. A maintainer following this pointer will search for a symbol that never existed.

packages/valory/skills/mech_interact_abci/states/final_states.py
```diff
@@ -56,3 +56,31 @@ class FinishedMechResponseRound(DegenerateRound):
...
+    """Offchain 402 path: structured 402, deposit multisend pending settlement.
+
+    Carries ``tx_submitter``+``most_voted_tx_hash`` for the consumer's
+    transaction-settlement skill; on ``PostTxSettlement`` the multiplexer
+    routes back to ``MechRequestRound`` to retry the original request_id
+    at the same nonce after the deposit lands.
+    """
+
+
+class FailedOffchainMechRequestRound(DegenerateRound):
```
OjusWiZard, line 79: **[Docs] Failure-cause list omits the whole `BAD_RESPONSE` category.** Malformed 402 body, non-200/402/503 status, unknown `paymentType`, or a failed `_validate_402_destination` cross-check all land here as `offchain_bad_response`, exactly like the enumerated timeout/503/over-cap/NVM cases — but none of them match this docstring, so an operator debugging that failure reason would conclude it's outside the documented design.

OjusWiZard (commented):
Multi-perspective automated review (correctness, silent failures, type design, test coverage, comments, simplification). Findings posted inline, ranked by severity: 2 Critical, 9 Important, plus doc/dead-code nits.

**Bottom line:** the one merge-gating question is the ERC-1271 / Safe-threshold finding on `_sign_request_id` — cheap to confirm against the PR #96 e2e setup (what threshold did the Safe have?). The logging/docstring/dead-code items could land as one follow-up commit on this branch; the poller test gap is test-only; the type-design items are reasonable to ticket.

What came back clean: the CID computation and EIP-712 `getRequestId`/domain-separator ports were verified byte-for-byte against the real `MechMarketplace` contract; `merge_extra_attributes` is used consistently on both paths; helper-level tests are unusually strong (golden EIP-712 vector, exact decoded-tx assertions, C1–C20 regression traceability, consensus-determinism test).

packages/valory/skills/mech_interact_abci/payloads.py
```diff
@@ -42,6 +42,17 @@ class MechRequestPayload(PrepareTxPayload):
     safe_contract_address: Optional[str]
     mech_requests: Optional[str]
     mech_responses: Optional[str]
+    # Offchain dispatch fields. None for the on-chain path (today's behaviour,
+    # unchanged). One of ``offchain_done|offchain_deposit_needed|
+    # offchain_all_failed`` when ``MechRequestBehaviour`` took the offchain
+    # HTTP path. ``MechRequestRound.end_block`` reads ``offchain_result`` to
+    # dispatch to the right ``Event`` variant; the other three carry the
```
OjusWiZard: **[Docs]** Miscount: "the other three" — after singling out `offchain_result`, only two fields remain (`offchain_pending_request`, `offchain_last_failure_reason`).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+    mech_url: str
+    sender: str
+    delivery_rate: int
+    ipfs_hash: str
+    ipfs_data: str
+    # Metadata UUID assigned by the caller (e.g. market-resolver's
+    # evaluate_answers). Downstream consumers correlate a response back to
+    # the request by matching MechInteractionResponse.nonce against this
+    # UUID -- same semantics as the legacy on-chain path
+    # (request.py:660: MechInteractionResponse(nonce=metadata.nonce, ...)).
```
OjusWiZard: **[Docs] Stale hard-coded line pointer.** The `MechInteractionResponse(nonce=metadata.nonce, ...)` construction is at `request.py:655-657`, not `:660` (line 660 is `return True`). Hard-coded line citations rot on any upstream edit — consider citing the function name (`_send_metadata_to_ipfs`) instead.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1835 @@
...
+    pending_request_json: Optional[str] = None
+    last_failure_reason: Optional[str] = None
+    # Set only on the deposit path. The Safe-multisend hash that the
+    # consumer's transaction_settlement skill submits before re-entering
+    # ``MechRequestRound`` to retry the POST.
+    tx_submitter: Optional[str] = None
+    tx_hash: Optional[str] = None
+
+
+_HEX_RE = re.compile(r"^0x[0-9a-fA-F]+$")
```
OjusWiZard: **[Dead code]** `_HEX_RE` is never referenced anywhere in the repo (verified via full grep). `_REQUEST_ID_HEX_RE` (line 498) is the one `PendingRequest.__post_init__` actually uses. Safe to delete.

packages/valory/skills/mech_interact_abci/rounds.py
```diff
@@ -208,4 +226,29 @@ class MechInteractAbciApp(AbciApp[Event]):
         },
         FinishedMechResponseRound: set(get_name(SynchronizedData.mech_responses)),
         FinishedMechResponseTimeoutRound: set(),
+        # Offchain happy path: the request POST got 200. ``mech_requests`` and
+        # ``mech_responses`` carry the request_id and the pending response
+        # placeholder so the response round can poll the offchain mech.
```
OjusWiZard: **[Docs] Comment omits the actual poll-target carrier.** It's `offchain_pending_request` — not `mech_requests`/`mech_responses` — that carries `mech_url`/`request_id` for `OffchainResponsePoller._load_pending` to poll. Worth naming it here (and it's also missing from the listed `db_pre_conditions`, which may be intentional since it's Optional, but the prose shouldn't imply the other two fields are sufficient).

OjusWiZard (dismissed):

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). `_retry_pending` now logs `outcome`+`status_code` on every non-DONE branch and routes DEPOSIT_NEEDED/OVER_CAP to `OFFCHAIN_402_INSUFFICIENT` while TIMEOUT/SERVER_BUSY/BAD_RESPONSE route via `_failure_label_for`. Regression test `test_retry_timeout_surfaces_timeout_not_insufficient` pins the fix.

packages/valory/skills/mech_interact_abci/states/base.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). The property now logs a warning on unparseable JSON (and on non-dict / wrong-type shapes), and the docstring documents both producers of this key (`_fresh_cycle`s DONE branch as well as the DEPOSIT_NEEDED branch) so the poller-side consumer is discoverable.

packages/valory/skills/mech_interact_abci/states/request.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). `MechRequestRound.end_block` now raises `ValueError` on a non-empty `offchain_result` that is not in `_OFFCHAIN_RESULT_TO_EVENT`, with a message listing the value and the expected set. Existing test that pinned the silent-fallthrough behaviour rewritten to assert the loud-fail (`test_unknown_offchain_result_raises`).

packages/valory/skills/mech_interact_abci/states/base.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). `merge_extra_attributes` now raises `ValueError` on reserved-key collisions instead of warn-and-clobber, so a caller substituting `tool` (or `prompt`/`nonce`/`schema_version`/`request_context`) via extras fails loudly at request-build time rather than silently mismatching what the mech settles against. Parametrised test covers each reserved key; benign extras still merge.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). The docstring now points at the real knob (`offchain_poll_timeout_seconds` in `models.py`) and the "module-level review notes" reference is dropped.

packages/valory/skills/mech_interact_abci/states/final_states.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). `FailedOffchainMechRequestRound` docstring now lists the BAD_RESPONSE category alongside the timeout / 503 / over-cap / NVM cases.

packages/valory/skills/mech_interact_abci/payloads.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). "the other three" → "the other two".

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). Line pointer replaced with a function-name reference (`_send_metadata_to_ipfs` in `request.py`) that will not rot on the next edit.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). `_HEX_RE` deleted.

packages/valory/skills/mech_interact_abci/rounds.py
DIvyaNautiyal07: Fixed in #110 (commit 7752486). Comment now names `offchain_pending_request` as the carrier of `mech_url` and `request_id` for `OffchainResponsePoller._load_pending`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Filed as #104 with the three-round FSM design (Prepare → Sign → Submit), the packer for Safe `checkSignatures`, and the EOA vs Safe branching. To confirm the concrete data on the PR#96 e2e question: it ran against a single-agent trader (`number_of_agents: 1` in `trader_pearl/service.yaml`), so the Safe is threshold-1 and the single-agent sig satisfies `checkSignatures` via ERC-1271. All currently shipped consumers of `mech_interact_abci` are single-agent by construction (trader, market-resolver, optimus) and all default `use_offchain: false`, so the failure path has zero shipped consumers today. Threshold > 1 support is the right issue but not a merge-gate for this rollup.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Filed as #105 with a suggested test list (targeted `_find_target_response` by primary/secondary requestId, `_apply_snapshot` in-place mutation assertion, `serialise_responses` round-trip, `run()` end-to-end fresh/resume/timeout/unknown-result). Agreed this is a real gap; tracked as its own follow-up so it can land with focused test-only diffs.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Filed as #109. Agreed the isinstance-based guard is the deliberate one and the drift reads as unintended; consolidating into a shared `load_pending_request(synced, logger)` in `states/base.py` is the right shape. Tracking as its own PR so the semantic change is reviewable independently of everything else.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Filed as #108. Good point that the per-agent HTTP polling has different consensus semantics than the on-chain event-log polling it replaces. Tracked with the three suggested mitigations (bounded NO_MAJORITY loops, delegate polling, synced-data deadline alignment) as options to pick between. Not a blocker for shipping since current single-agent Pearl consumers cannot exercise the divergence, but worth the deliberate design pass before a multi-agent consumer opts in.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Filed as #107 (bundled with the `Optional[PaymentChallenge]` narrowing item since both are one refactor pass on the executor / poller collaborator surface). Agreed the `Any` typing erases mypy coverage and duplicating the `__init__` between executor and poller is real drift risk; extracting a small typed shared base fixes both in one change.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Filed as #107 (bundled with the `Any`-typed-collaborator item; both fit one type-design refactor pass on the executor). Structural modelling of the outcome variants (per-outcome carrier or tagged union with a non-Optional challenge) is the direction; assert-narrow is the quick alternative if the tagged-union approach is heavier than warranted.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Filed as #106 with the enumerated executor-branch gaps (NVM 402 short-circuit, `_pick_next_mech` per-mech URL fallback + dynamic-selection gate, `_resolve_payment_type` invalid-length/type branches, `_resolve_chain_id_int` / `_read_on_chain_nonce` STATE-with-bad-value branches, `run()` fresh-vs-resume dispatch). Tracked as its own follow-up so the tests land with focused stub extensions.

OjusWiZard (approved):

packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -632,13 +637,8 @@ def _send_metadata_to_ipfs(  # pragma: no cover
...
         payload = asdict(metadata)
-        extras = payload.pop("extra_attributes", None) or {}
-        clobbered = extras.keys() & payload.keys()
-        if clobbered:
-            self.context.logger.warning(
-                f"extra_attributes override reserved request keys: {clobbered}"
-            )
-        payload.update(extras)
+        extras = payload.pop("extra_attributes", None)
+        merge_extra_attributes(payload, extras, self.context.logger)
```
bennyjo, line 641: This also changes the on-chain path: colliding extras used to log a warning and continue, now they raise. A consumer that passes a reserved key in `extra_attributes` would start crashing on every request. Have we checked no consumer does this?

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1861 @@
...
+            mech_address=mech_address.lower(),
+            mech_url=mech_url,
+            sender=self._safe_address(),
+            delivery_rate=delivery_rate,
+            ipfs_hash=ipfs_hash,
+            ipfs_data=ipfs_data,
+            metadata_nonce=metadata_nonce,
+        )
+
+    def _post_signed_request(
```
bennyjo, line 1039: On a multi-agent service, every agent POSTs the same `request_id`. If the mech rejects duplicates instead of returning 200, agents get different outcomes and the round loops `NO_MAJORITY` forever. Does the server accept a re-submitted `request_id`?

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,380 @@
...
+
+            if status_code == 404:
+                self._logger.error(
+                    f"Offchain poll {url} returned 404 for request_id "
+                    f"{request_id_int_str}; the mech does not know about "
+                    "this request (likely restarted). Fast-failing."
+                )
+                return _PollSnapshot(status="rejected", error="not_found")
+
+            if status_code is not None and 500 <= status_code < 600:
```
bennyjo: The framework returns `status_code=600` on connection failures instead of raising, so a dead mech never triggers this fast-fail — each 600 resets the counter and the loop runs the full 300s. Same gap in `_post_signed_request`, where 600 shows up as `OFFCHAIN_BAD_RESPONSE` instead of `TIMEOUT`.

packages/valory/skills/mech_interact_abci/skill.yaml
```diff
@@ -215,6 +222,10 @@ models:
         response_timeout: 300
         use_offchain: false
         offchain_url: null
+        auto_deposit_cap_per_cycle: null
+        offchain_poll_interval_seconds: 3.0
+        offchain_poll_timeout_seconds: 300.0
+        offchain_failover_max_retries: 2
```
bennyjo, line 228: `offchain_deposit_target_calls` is missing here, and service overrides are validated against this block — so operators can't actually tune it. Worth adding before the downstream hash bumps.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
```diff
@@ -136,12 +136,47 @@ def populate_tools(
...
+            # Operators inject it via ``make update-metadata`` in the
+            # mech-deployments repo (see its README). Manifests that predate
+            # the offchain rollout will not carry the key; the offchain
+            # behaviour falls back to ``MechMarketplaceConfig.offchain_url``
+            # for those mechs.
+            manifest_url = self._extract_manifest_url(res_raw)
             for mech in mechs:
                 mech.relevant_tools |= metadata_tools
+                if manifest_url is not None:
+                    mech.http_url = manifest_url
```
bennyjo, line 151: This runs with `use_offchain=false` too, so every serialized `mechs_info` blob now carries an `http_url` key. Harmless at runtime (deterministic, parse is guarded) — flagging so the on-chain shape change is a known one, not an accident.

packages/valory/skills/mech_interact_abci/behaviours/request.py
DIvyaNautiyal07: Grepped every downstream consumer that still uses this skill (trader, optimus, market-creator, meme-ooorr, IEKit) for `MechMetadata(extra_attributes=...)` and for direct `extra_attributes={`. Zero production sites pass any of the reserved keys (`prompt` / `tool` / `nonce` / `schema_version` / `request_context`) — the only extras seen in the wild are `topics`, `num_questions`, `max_tokens`, `system`, and the mech-tools-side test fixtures use the same shape. Only tests reference `extra_attributes` at all; production `MechMetadata` construction in each of those repos passes just the four required fields. Safe to keep the raise. Addressed in commit d71c1a0 on this branch.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Checked mech-side. `_enqueue_offchain_request` in `mech/packages/valory/skills/task_execution/handlers.py` unconditionally appends to `pending_tasks` and overwrites `in_memory_requests[request_id]` with no duplicate check, and `_send_ok_response` fires 200 in every branch that doesn't fail auth / balance. So all N agents on a multi-agent service get the same 200 outcome for the same signed request_id, threshold reaches, no NO_MAJORITY loop. Addressed in commit d71c1a0 on this branch.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in commit d71c1a0 on this branch. `valory/http_client/connection.py:113` confirms 600 is the AEA synthetic connection-failure code. Two paired changes:

- **Poller**: extended the fast-fail range from `500 <= status_code < 600` to `500 <= status_code < 600 or status_code == 600`, so five consecutive connection failures surface `server_unavailable` on the same budget as five 5xx (instead of spinning the full 300s).
- **Executor**: added an explicit `status == 600` branch returning `OffchainAttemptOutcome.TIMEOUT` (was `BAD_RESPONSE`), so the failover loop tries the next ranked mech instead of terminating with `OFFCHAIN_BAD_RESPONSE` on the first attempt.

Regression tests: `600 -> TIMEOUT` added to `test_status_to_outcome`; `5 * 600 -> server_unavailable` added to the poller fast-fail suite.

packages/valory/skills/mech_interact_abci/skill.yaml
DIvyaNautiyal07: Fixed in commit d71c1a0 on this branch. Added `offchain_deposit_target_calls: 10` to the `mech_marketplace_config` args block in `skill.yaml` (default matches `models.py:261`). Service overrides via the standard args-validation path now work.

packages/valory/skills/mech_interact_abci/behaviours/mech_info.py
DIvyaNautiyal07: Acknowledged in commit d71c1a0 on this branch. The shape change is intentional: `populate_tools` runs the same manifest fetch under both flag states, and hydrating `http_url` unconditionally keeps the URL discovery a single code path (deterministic on identical input, parse-guarded on the JSON). The alternative — gating on `use_offchain` here — would create a subtle mode-dependent `mechs_info` shape that consumers reading via the on-chain path could still be surprised by after a flag flip. Flagging it in the PR description so it's on the record as a known on-chain-shape byproduct rather than an accident.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1872 @@
...
+            numRequests=1,
+            nonce=metadata_nonce,
+            result=None,
+            error="Unknown",
+        )
+        return json.dumps([dataclasses.asdict(placeholder)], ensure_ascii=True)
+
+    # ---------- state-on-the-wire helpers ----------------------------------
+
+    def _load_pending_request(self) -> Optional[PendingRequest]:
```
bennyjo, line 1854: `offchain_pending_request` is persisted on the DONE path but never cleared afterwards, so a second offchain request in the same period would resume the already-delivered one and silently drop the new prompt — worse, its answer gets paired with the new request's metadata. Would a guard here comparing `pending.metadata_nonce` against the current `mech_requests[0].nonce` (discard on mismatch) work?

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1872 @@
...
+        # exactly one entry per cycle; the offchain wire shape supports one
+        # request per ``request_id`` so we serialise on the first entry and
+        # leave multi-request batching for a follow-up.
+        requests = self._synced.mech_requests
+        if not requests:
+            self._logger.info("Offchain branch reached with no mech_requests; skipping")
+            return OffchainCycleResult(offchain_result=Event.OFFCHAIN_DONE.value)
+
+        request_meta: MechMetadata = requests[0]
+        ipfs_hash, _, ipfs_data = build_request_metadata(
```
bennyjo: `compute_cidv1_bytes` raises `ValueError` for metadata over 256 KiB and nothing on this path catches it, so one oversized prompt becomes a crash loop — the same `mech_requests` re-enter the round each period and raise again. Suggest catching it here and returning `OFFCHAIN_ALL_FAILED` so it surfaces as a clean failure (a prompt that size works fine on the on-chain IPFS path today).

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1872 @@
...
+        # ``hash_payload_to_hex`` expects a raw 64-char hex (no ``0x``
+        # prefix) for ``safe_tx_hash`` and a ``bytes`` ``data`` (it calls
+        # ``.hex()`` internally); base.py:116 stores the same form via
+        # ``safe_hash[2:]`` and feeds ``self.multisend_data`` (bytes).
+        return hash_payload_to_hex(
+            safe_tx_hash[2:],
+            value,
+            SAFE_GAS,
+            to_address,
+            bytes(data) if not isinstance(data, bytes) else data,
```
bennyjo: `bytes(data) if not isinstance(data, bytes) else data` is a no-op for `HexBytes` (it's a `bytes` subclass), so on hexbytes 0.3.x — still allowed by tox.ini's range — the token path packs a literal `0x` into the payload and settlement-side decode raises, failing every token-402 deposit. Passing `bytes(data)` unconditionally fixes it; also worth adding a `skill_input_hex_to_payload` round-trip assertion to `test_token_path_matching_destination_proceeds_to_build` (test_offchain_request_helpers.py:1932), which currently only checks `tx_hash is not None` — the native test has this, the token/DELEGATE_CALL path doesn't.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
```diff
@@ -0,0 +1,1872 @@
...
+            mech_address=mech_address.lower(),
+            mech_url=mech_url,
+            sender=self._safe_address(),
+            delivery_rate=delivery_rate,
+            ipfs_hash=ipfs_hash,
+            ipfs_data=ipfs_data,
+            metadata_nonce=metadata_nonce,
+        )
+
+    def _post_signed_request(
```
bennyjo, line 1039: Every agent in an n>1 service POSTs the same request_id independently, and the mech server doesn't dedupe — it re-enqueues the task and a straggler POST even discards an already-completed response (`task_execution/handlers.py` pops the stored result on a duplicate). Is n>1 explicitly out of scope for now? Happy to leave as-is if so — flagging because it's silent (every POST returns 200) and multiplies tool cost.

packages/valory/skills/mech_interact_abci/states/base.py
```diff
@@ -87,14 +100,60 @@ class MechMetadata:
...
+    if clobbered:
+        if logger is not None:
+            logger.error(
+                "extra_attributes contains reserved request keys: %s", clobbered
+            )
+        raise ValueError(
```
bennyjo, line 154: This changes the on-chain path too: `_send_metadata_to_ipfs` previously warned and clobbered on reserved-key overlap, now it raises — a consumer passing an overlapping `extra_attributes` key goes from a log line to a request behaviour that fails on every act. Is that intentional? If so, worth a note in the PR description since it contradicts the "on-chain path unchanged" framing.

packages/valory/skills/mech_interact_abci/behaviours/request.py
```diff
@@ -1003,9 +1003,60 @@ def _build_skip_payload(self) -> MechRequestPayload:
...
+        )
+
     def async_act(self) -> Generator:  # pragma: no cover
         """Do the action."""
 
+        # ``is True`` not bool() so unrelated truthy mocks in unit tests
+        # never accidentally take the offchain branch; the config validator
+        # already enforces the value is a real bool when ``use_offchain``
+        # is enabled at deploy time.
+        if self.mech_marketplace_config.use_offchain is True:
```
bennyjo, line 1056: The comment says the validator enforces a real bool, but `MechMarketplaceConfig.__post_init__` only truthiness-checks — a string `"true"` (e.g. from an env-var override) passes offchain validation, forcing the cap to be configured, yet fails `is True` here and silently runs the on-chain path. Suggest either coercing/type-checking in the validator or gating on truthiness here.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,392 @@
...
+                # from a real tool failure, so the truncated-payload shape
+                # would keep looking like a bad prediction. Flag it.
+                self._logger.warning(
+                    "Offchain 'ok' status but 'response' envelope "
+                    "missing/malformed: %r. Treating as a failed poll.",
+                    envelope,
+                )
+                inner_result = envelope
+            return _PollSnapshot(
+                status="ok",
```
bennyjo: The drift branches log "Treating as a failed poll" but still return `status="ok"` with `result=None`, which downstream records as `error="Unknown"` — indistinguishable from a routine tool failure. If drift should be distinguishable, return a distinct status/error here; if not, the log message should stop claiming it's treated as a failure.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
```diff
@@ -0,0 +1,392 @@
...
+                self._logger.warning(
+                    f"Offchain poll {url} raised {exc.__class__.__name__}: {exc}"
+                )
+                response = None
+
+            status_code = (
+                getattr(response, "status_code", None) if response is not None else None
+            )
+
+            if status_code == 404:
```
bennyjo, line 207: One 404 permanently rejects a paid-for request, while 5xx gets a 5-consecutive tolerance — if the mech registers the request asynchronously after its POST 200, a racy first GET kills it. Is registration synchronous on the server side? If not, giving 404 the same consecutive-failure tolerance seems safer.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 0da77e2. `_load_pending_request` now compares the pending `metadata_nonce` against `mech_requests[0].nonce` (your suggested guard) and discards on mismatch — the next period starts a fresh cycle instead of re-POSTing the delivered request. Regression tests cover both the mismatch-discard path and the same-nonce resume path so the deposit-just-settled retry leg keeps working.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 0da77e2. Wrapped `build_request_metadata` in try/except and route the oversize ValueError through `OFFCHAIN_ALL_FAILED` with a new `OFFCHAIN_METADATA_OVERSIZE` last-failure reason. Regression test asserts no HTTP POST is attempted when the metadata is oversized, so the FSM surfaces the failure cleanly instead of re-crashing every period.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Fixed in 0da77e2. Now passes `bytes(data)` unconditionally — since HexBytes is a bytes subclass the isinstance guard was skipping the coercion, and on older hexbytes<1 versions still in the tox range the stringification-downstream path would prepend `0x` into the payload. Removed the guard so both hexbytes lines produce identical bytes going into `hash_payload_to_hex`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_request.py
DIvyaNautiyal07: Confirming n>1 is out of scope for this PR. You're right about the straggler pop — `task_execution/handlers.py` pops any stored response for a duplicate `request_id`, so under n>1 a late-arriving agent POST silently throws away a completed response and the next FSM period re-processes the same request. It converges (all agents eventually see the same result) but wastes LLM work and adds latency. Filing as a follow-up to re-scope when we ship multi-agent — the fix will need mech-side dedupe (or an idempotency window), not just client-side coordination.

packages/valory/skills/mech_interact_abci/states/base.py
DIvyaNautiyal07: Confirming intentional and noting in the PR description. The reserved-key overlap raise change was deliberate: the previous warn-and-clobber path could silently substitute a mech tool via a rogue `extra_attributes.tool` (or drop the prompt / nonce), and grepping every downstream consumer (trader / optimus / market-creator / meme-ooorr / IEKit) turned up zero production sites that pass any of `{prompt, tool, nonce, schema_version, request_context}` as extras — only tests, and their extras are all non-reserved (`topics`, `num_questions`, `max_tokens`, `system`). So the on-chain path change is real but silent to consumers today. Adding a note to the PR description as suggested.

packages/valory/skills/mech_interact_abci/behaviours/request.py
DIvyaNautiyal07: Fixed in 0da77e2 (validator side). Added an `isinstance(use_offchain, bool)` check in `MechMarketplaceConfig.__post_init__` so a truthy non-bool (e.g. env-var string "true") fails loud at startup instead of passing cap-required validation and then silently taking the on-chain path via the `is True` dispatch guard. Kept the `is True` guard in `async_act` (removes the ambiguity for the test suite too). Parametrised regression covers `"true"` / `"false"` / `1` / `0` / `"yes"` / `None`.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in 0da77e2. Drift branches now return `status="rejected"` with a distinct `error="envelope_drift"`, so `_apply_snapshot` sets `target.error = "envelope_drift"` on the response record — trader's `DecisionReceive` can tell a mech-side schema mismatch apart from a routine tool failure (which still lands as `error="Unknown"`). Covers both the dict-missing-`result` case and the non-dict `response` field case symmetrically. Updated the existing missing-result test + added a new malformed-envelope test to lock the behaviour.

packages/valory/skills/mech_interact_abci/behaviours/offchain_response.py
DIvyaNautiyal07: Fixed in 0da77e2. Async registration is synchronous on the current server (POST 200 returns after `pending_tasks.append`), so the specific race isn't present today — but the LB / rolling-deploy blip case is real, and the fix costs almost nothing. 404 now shares the same `_MAX_CONSECUTIVE_5XX` tolerance as 5xx/600, with its own counter so 404s and server-failures don't cross-reset. A single 404 no longer fast-fails; five consecutive still surface as `"not_found"`. Also renamed the poller counter to `consecutive_failures` since it now buckets 5xx/600 together. Regression tests cover both the single-404 recover path and the five-in-a-row fast-fail path.

